PR #26, opened 2020-12-05
Add Key alias in NestedSpanAttributesProtocol

### LINKED ISSUE #10: Introduce Key alias in nested span attributes protocol
now:

```
public struct NestedSpanAttributes: NestedSpanAttributesProtocol {
    public var transport: SpanAttributeKey<String> { .init(name: SpanAttributeName.Net.transport) }
```

better:

```
public struct NestedSpanAttributes: NestedSpanAttributesProtocol {
    public var transport: Key<String> { .init(name: SpanAttributeName.Net.transport) }
```

so let's add a typealias in `NestedSpanAttributesProtocol`

slashmo:
As described in #10, this makes for a slightly nicer API when defining `SpanAttribute` constants.

- Closes #10

### DIFF
--- a/Sources/Tracing/Span.swift
+++ b/Sources/Tracing/Span.swift
@@ -169,6 +169,10 @@ extension NestedSpanAttributesProtocol {
     }
 }
 
+extension NestedSpanAttributesProtocol {
+    public typealias Key = SpanAttributeKey
+}
+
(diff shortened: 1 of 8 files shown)


ktoso (approved):
Nice thanks 👍

PR #31, opened 2020-12-14
Replace SpanStatus.CanonicalCode with OTel status code

### LINKED ISSUE #29: Remove SpanStatus CanonicalCode
OpenTelemetry deprecated GRPC-like status codes a while ago. We might want to remove them from our API as well:

**Current spec**
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-status

**Change in `.proto` file**
https://github.com/open-telemetry/opentelemetry-proto/commit/59c488bfb8fb6d0458ad6425758b70259ff4a2bd#diff-5c407ac9c675a60170f94ab620f5329f25e5388d6790864b7799abce66f79332

*This commit also added backwards-compatibility notes we might want to follow.*

slashmo:
Closes #29

### DIFF
--- a/Sources/Tracing/Span.swift
+++ b/Sources/Tracing/Span.swift
@@ -589,6 +589,7 @@ public struct SpanStatus {
     public let message: String?
 
     /// Create a new `SpanStatus`.
+    ///
     /// - Parameters:
     ///   - code: The `Code` of this `SpanStatus`.
     ///   - message: The optional descriptive message of this `SpanStatus`. Defaults to nil.
@@ -601,8 +602,6 @@ public struct SpanStatus {
     ///
     /// - SeeAlso: For the semantics of status codes see [OpenTelemetry Specification: setStatus](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-status)
     public enum Code {
-        /// The default status.
-        case unset
         /// The Span has been validated by an Application developer or Operator to have completed successfully.
         case ok
         /// The Span contains an error.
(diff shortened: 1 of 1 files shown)


Sources/Tracing/Span.swift
```diff
@@ -583,61 +583,29 @@ extension SpanAttributes: ExpressibleByDictionaryLiteral {
 // ==== ----------------------------------------------------------------------------------------------------------------
 // MARK: Span Status
 
-/// Represents the status of a finished Span. It's composed of a canonical code in conjunction with an optional descriptive message.
+/// Represents the status of a finished Span. It's composed of a status code in conjunction with an optional descriptive message.
 public struct SpanStatus {
-    public let canonicalCode: CanonicalCode
+    public let code: Code
```
ktoso, line 588: Should it be possible to set one with undefined status code?

Sources/Tracing/Span.swift
```diff
@@ -583,61 +583,29 @@ extension SpanAttributes: ExpressibleByDictionaryLiteral {
...
-        /// Operation is not implemented or not supported/enabled in this service.
-        case unimplemented
-        /// Internal errors. Means some invariants expected by underlying system has been broken.
-        case `internal`
-        /// The service is currently unavailable. This is a most likely a transient condition and may be corrected by retrying with a backoff.
-        case unavailable
-        /// Unrecoverable data loss or corruption.
-        case dataLoss
-        /// The request does not have valid authentication credentials for the operation.
-        case unauthenticated
```
ktoso, line 640: Good riddance, was quite complex 👍

Sources/Tracing/Span.swift
slashmo: > Should it be possible to set one with undefined status code?
> @ktoso 

Good question ... I don't think so, as unset basically means you haven't altered the status. The only use-case would be to set it at first, then reset it afterwards. But I don't think that's a real concern.

Sources/Tracing/Span.swift
slashmo: Agreed 😊

ktoso (approved):

PR #33, opened 2020-12-14
Call span.end() in the happy path in withSpan

### LINKED ISSUE #30: withSpan() doesn’t work as documented
The documentation comments clearly state:

> DO NOT `end()` the passed in span manually. It will be ended automatically when the `function` returns.

However, there is no `end()` on the happy path, only if an error occurs.

Of course I could make a PR, but when I encountered this I convinced myself that such an obvious omission _might_ be intentional. :-)

JensAyton:
Updates behaviour of `withSpan` to match documentation.

I would consider also making the `context` form of `withSpan` (which is untested) call into the `baggage` variant.

Fixes #30

### DIFF
--- a/Tests/TracingTests/TracerTests.swift
+++ b/Tests/TracingTests/TracerTests.swift
@@ -71,7 +71,7 @@ final class TracerTests: XCTestCase {
         }
 
         XCTAssertEqual(value, "yes")
-        XCTAssertEqual(spanEnded, true)
+        XCTAssertTrue(spanEnded)
     }
 
     func testWithSpan_throws() {
@@ -89,7 +89,7 @@ final class TracerTests: XCTestCase {
                 throw ExampleSpanError()
             }
         } catch {
-            XCTAssertEqual(spanEnded, true)
+            XCTAssertTrue(spanEnded, true)
             XCTAssertEqual(error as? ExampleSpanError, ExampleSpanError())
             return
         }
--- a/Tests/TracingTests/TracerTests.swift
+++ b/Tests/TracingTests/TracerTests.swift
@@ -63,11 +63,15 @@ final class TracerTests: XCTestCase {
             InstrumentationSystem.bootstrapInternal(NoOpTracer())
         }
 
+        var spanEnded = false
+        tracer.onEndSpan = { _ in spanEnded = true }
+
         let value = tracer.withSpan("hello", baggage: .topLevel) { _ in
             "yes"
         }
 
         XCTAssertEqual(value, "yes")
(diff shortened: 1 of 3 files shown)


Tests/TracingTests/TracerTests.swift
```diff
@@ -63,11 +63,15 @@ final class TracerTests: XCTestCase {
...
 
+        var spanEnded = false
+        tracer.onEndSpan = { _ in spanEnded = true }
+
         let value = tracer.withSpan("hello", baggage: .topLevel) { _ in
             "yes"
         }
 
         XCTAssertEqual(value, "yes")
+        XCTAssertEqual(spanEnded, true)
```
slashmo, line 74:
```suggestion
        XCTAssert(spanEnded)
```

Tests/TracingTests/TracerTests.swift
```diff
@@ -77,11 +81,15 @@ final class TracerTests: XCTestCase {
...
 
+        var spanEnded = false
+        tracer.onEndSpan = { _ in spanEnded = true }
+
         do {
             _ = try tracer.withSpan("hello", baggage: .topLevel) { _ in
                 throw ExampleSpanError()
             }
         } catch {
+            XCTAssertEqual(spanEnded, true)
```
slashmo, line 92:
```suggestion
            XCTAssert(spanEnded)
```

slashmo (approved):
Thanks for taking this on, @JensAyton! 👍 LGTM, just some minor nitpicks to clean up the assertions.

slashmo:
> I would consider also making the context form of withSpan (which is untested) call into the baggage variant.
> @JensAyton

Sounds good to me. Would you mind adding this as well?

tomerd:
@swift-server-bot add to whitelist

Tests/TracingTests/TracerTests.swift
tomerd: also `XCTAssertTrue` if you want to be more explicit

ktoso:
```suggestion
        XCTAssertTrue(spanEnded)
```

Tests/TracingTests/TracerTests.swift
ktoso:
```suggestion
            XCTAssertTrue(spanEnded, true)
```

ktoso:
Let me merge this and let's do 

> I would consider also making the context form of withSpan (which is untested) call into the baggage variant.


in a follow up please @JensAyton ?

Thanks a lot!

PR #41, opened 2021-01-27
Remove UseCases dir from scripts

slashmo:
The `UseCases` directory has been removed a while ago, but it was still included in some scripts.


ktoso (approved):
Good cleanup :)

Issue #45, opened 2021-03-14
Span.recordError should take optional SpanAttributes


slashmo:
[The OTel specification states](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception) that Span’s `recordException` (in our case `recordError`) should take an additional argument to optionally specify `SpanAttributes`. These additional attributes (in case of OTel) are then merged into the resulting event’s attributes.

> If RecordException is provided, the method MUST accept an optional parameter to provide any additional event attributes (this SHOULD be done in the same way as for the AddEvent method). If attributes with the same name would be generated by the method already, the additional attributes take precedence.

ktoso:
Good catch!

PR #46, opened 2021-03-29
Conform SpanStatus to Equatable

slashmo:
It would be nice to have `SpanStatus` conform to `Equatable` for easier testing.


ktoso (approved):
makes sense to me 👍 

Would you want a patch release?

slashmo:
> Would you want a patch release?
> @ktoso 

No hurry on my end. I think we can wait until #45 is closed (which I'll work on soon), and combine both in a patch release.

PR #49, opened 2021-04-19
Add async versions of `withSpan` API

slashmo:
Adds support for creating spans in an async context with automatic "baggage propagation" through task locals.

### DIFF
--- a/Sources/Tracing/Tracer.swift
+++ b/Sources/Tracing/Tracer.swift
@@ -127,3 +127,65 @@ extension Tracer {
         }
     }
 }
+
+#if compiler(>=5.5) // we cannot write this on one line with `&&` because Swift 5.0 doesn't like it...
+#if compiler(>=5.5) && $AsyncAwait
+import _Concurrency
+
+@available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *)
+extension Tracer {
+    /// Execute a specific async task within a newly created `Span`.
+    ///
+    /// DO NOT `end()` the passed in span manually. It will be ended automatically when the `function` completes.
+    ///
+    /// - Parameters:
+    ///   - name: The name of the operation being traced. This may be a handler function, database call, ...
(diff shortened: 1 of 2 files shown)


Sources/Tracing/Tracer.swift
```diff
@@ -127,3 +127,65 @@ extension Tracer {
         }
     }
 }
+
+#if compiler(>=5.5) // we cannot write this on one line with `&&` because Swift 5.0 doesn't like it...
```
ktoso, line 131: oh, lol okey :)

slashmo: Copied directly from @weissi's NIO async/await support 😁

weissi: do you support Swift 5.0 in this repo? If no, then you can remove this extra line, from Swift 5.1 on `#if compiler(>=5.5) && $AsyncAwait` works as expected. In Swift 5.0 that line will fail compilation and error (think it complains about something that has to be numeric or so).

ktoso (approved):
Fantastic, this'll be great 👍

Sources/Tracing/Tracer.swift
ktoso: For better or worse we currently do support 5.0, yes.

Though to be honest we'll drop it all and require async soon I guess...?
But for now yes, we do.

slashmo: Yeah, I don't think we need to maintain backwards compatibility here.

weissi: Given that this repo isn't 1.0.0 tagged, I'd recommend dropping all old Swift versions :P

ktoso: Yeah, we'll do that.

The discussions last year during gsoc were on the dramatic side with people requesting 5.0... 😉 

Now as we want to require task-locals we'll rehash the lib and require async features anyway and drop the 5.0, not much to be discussed about it this year about it.

slashmo:
Closed in favor of #52

PR #52, opened 2021-08-10
Use dependency-free Baggage module

### LINKED ISSUE #51: Update documentation to account for automatic context propagation
We should update the documentation to include a section about automatic context propagation, and remove any mention of the soon to be removed `LoggingContext`.

### LINKED ISSUE #53: Move OpenTelemetryTracingSupport into swift-distributed-tracing-extras
The constants for using OTel spec compliant span attributes should instead life in a separate swift-distributed-tracing-extras  package.

### LINKED ISSUE #54: Support automatic context propagation through task locals
- [x] https://github.com/apple/swift-distributed-tracing-baggage/pull/13
- [x] #55

### LINKED ISSUE #55: Add `withSpan` overloads for automatic context propagation
We should add `withSpan` overloads which take no `Baggage` argument, and instead use the task local `Baggage.current`. We'll need one synchronous and one async version (at least until `reasync` was added to the language). These overloads should execute the given operation within a child task where `Baggage.$current` is bound to the newly created span's `Baggage`. That way spans started within that operation will automatically become children of the outer span (depending on the tracer implementation).

slashmo:
- Closes #53
- Closes #54
- Closes #55


slashmo:
@tomerd Regarding CI we'd need the same update as for the Baggage repo. Remove 5.0 & 5.1 and add 5.4 & 5.5 instead.

Sources/Tracing/Tracer.swift
```diff
@@ -58,72 +59,92 @@ extension Tracer {
...
 // ==== ----------------------------------------------------------------------------------------------------------------
-// MARK: Starting spans: `withSpan`
+// MARK: Starting spans: Task-local Baggage propagation
 
+#if swift(>=5.5)
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
 extension Tracer {
-    /// Execute a specific task within a newly created `Span`.
+    /// Execute the given operation within a newly created `Span`,
+    /// started as a child of the currently stored task local `Baggage.current` or as a root span if `nil`.
```
ktoso, line 100: excellent :)

Sources/Tracing/Tracer.swift
```diff
@@ -58,72 +59,92 @@ extension Tracer {
...
     public func withSpan<T>(
         _ operationName: String,
-        baggage: Baggage,
         ofKind kind: SpanKind = .internal,
-        _ function: (Span) throws -> T
-    ) rethrows -> T {
-        let span = self.startSpan(operationName, baggage: baggage, ofKind: kind)
+        _ operation: Span async throws -> T
+    ) async rethrows -> T {
+        let span = self.startSpan(operationName, baggage: .current ?? .topLevel, ofKind: kind)
```
ktoso, line 138: 👍

Sources/Tracing/Tracer.swift
```diff
@@ -58,72 +59,92 @@ extension Tracer {
...
         _ operationName: String,
-        context: LoggingContext,
-        ofKind kind: SpanKind = .internal
-    ) -> Span {
-        return self.startSpan(operationName, baggage: context.baggage, ofKind: kind, at: .now())
+        baggage: Baggage,
+        ofKind kind: SpanKind = .internal,
+        _ operation: (Span) throws -> T
+    ) rethrows -> T {
+        let span = self.startSpan(operationName, baggage: baggage, ofKind: kind)
```
ktoso, line 82: should we add some docs on startSpan that "prefer `withSpan` whenever possible as it automatically manages the spans lifetime; startSpan is provided for APIs which are not able to carry the span from beginning to end enclosed within a closure (e.g. if a span is started in one callback invoked by a 3rd party, and ended in another)."

?

Sources/Tracing/Tracer.swift
```diff
@@ -58,72 +59,92 @@ extension Tracer {
...
 
 // ==== ----------------------------------------------------------------------------------------------------------------
-// MARK: Span creation: with `LoggingContext`
+// MARK: Starting spans: `withSpan`
 
 extension Tracer {
-    /// Start a new `Span` with the given `Baggage` starting at `DispatchWallTime.now()`.
+    /// Execute a specific task within a newly created `Span`.
+    ///
+    /// DO NOT `end()` the passed in span manually. It will be ended automatically when the `operation` returns.
```
ktoso, line 67: 👍

ktoso (approved):
This is looking good! I'll get CI sorted out once California wakes up and I can poke some folks :)

slashmo:
@yim-lee Looks like in this repo 5.4 & 5.5 are not yet required to pass by the branch protection rules. Could you please add that in as well?

ktoso:
Seems to me we could merge this?

slashmo:
The 5.5 CI failed due to the Ruby version not being high enough to install Jazzy:
> ERROR:  Error installing jazzy:
	The last version of jazzy (>= 0) to support your Ruby & RubyGems was 0.13.7. Try installing it with `gem install jazzy -v 0.13.7`
	jazzy requires Ruby version >= 2.6.3. The current ruby version is 2.5.0.

CC @ktoso

Package.swift
```diff
@@ -8,7 +8,7 @@ let package = Package(
         .library(name: "Tracing", targets: ["Tracing"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-distributed-tracing-baggage.git", from: "0.2.0"),
+        .package(url: "https://github.com/apple/swift-distributed-tracing-baggage.git", .upToNextMinor(from: "0.2.0")),
```
ktoso, line 11: I don't think we should do this in a library.

The current issue is just an outcome that we're removing/breaking Apis because all this is pre 1.x, it's not something we need to future guard against.

ktoso:
very weird with that docker issue... I'll look into it with @yim-lee in the morning

yim-lee:
> @yim-lee Looks like in this repo 5.4 & 5.5 are not yet required to pass by the branch protection rules. Could you please add that in as well?

Sorry I missed this. I've made 5.4 and 5.5 required as well. FYI branch rules can be updated by any GH repo admin. 

> The 5.5 CI failed due to the Ruby version not being high enough to install Jazzy:

This is not specific to Swift 5.5 actually. One can reproduce it locally if they build the docker image with `--no-cache` I believe (at least I can).

To fix the problem, either do what the error message suggests and update `docker/Dockerfile` to do `gem install jazzy -v 0.13.7`, or modify it to use a newer Ruby version.

In case it helps, this mimics what CI does:

```
docker-compose -f docker/docker-compose.yaml -f docker/docker-compose.1804.55.yaml build
```

slashmo:
@yim-lee Thanks! Specifying the version explicitly works 👍

PR #60, opened 2022-01-29
Update to Sendable Baggage

ktoso:


slashmo (approved):
LGTM

slashmo:
@ktoso This will be tagged as `0.3.0` though, right?

ktoso:
Yeah this would be 0.3 of tracing -- that works right?
Also for your Tracer impls

slashmo:
> Yeah this would be 0.3 of tracing -- that works right?
> Also for your Tracer impls

Yep, OpenTelemetry Swift already supports `Sendable`, so for me personally it doesn't make a difference if this is gonna be tagged as `0.2.3` or `0.3.0`, but we should still do `0.3.0` as the experience might differ for others and it's nice to have release versions of the Baggage and Tracing repos aligned.

ktoso:
yeah it makes sense to tag them in unison

PR #64, opened 2022-08-19
+span additional withSpan to ease entering async world from non-async world

ktoso:
I'm wondering if we need to offer this.

Just an idea for now, but I've hit this when calling swift code from C, where we could not have had set the task-local, so here we're "entering the async world". We could set the `Baggage` around the withSpan, but that means two levels of nesting and one task local set more than we actually need.

### DIFF
--- a/Sources/Tracing/Tracer.swift
+++ b/Sources/Tracing/Tracer.swift
@@ -150,5 +150,36 @@ extension Tracer {
             throw error // rethrow
         }
     }
+
+    /// Execute the given async operation within a newly created `Span`,
+    /// started as a child of the passed in `Baggage` or as a root span if `nil`.
+    ///
+    /// DO NOT `end()` the passed in span manually. It will be ended automatically when the `operation` returns.
+    ///
+    /// - Parameters:
+    ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
+    ///   - baggage: The baggage to be used for the newly created span. It may be obtained by the user manually from the `Baggage.current`,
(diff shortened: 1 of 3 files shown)


Sources/Tracing/Tracer.swift
```diff
@@ -150,5 +150,36 @@ extension Tracer {
...
+    ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
+    ///   - baggage: The baggage to be used for the newly created span. It may be obtained by the user manually from the `Baggage.current`,
+    //               task local and modified before passing into this function. The baggage will be made the current task-local baggage for the duration of the `operation`.
+    ///   - kind: The `SpanKind` of the `Span` to be created. Defaults to `.internal`.
+    ///   - operation: operation to wrap in a span start/end and execute immediately
+    /// - Returns: the value returned by `operation`
+    /// - Throws: the error the `operation` has thrown (if any)
+    public func withSpan<T>(
+        _ operationName: String,
+        baggage: Baggage?,
```
slashmo, line 169: To keep it more consistent with the synchronous counterpart of this `withSpan`, I'd make `baggage` a non-optional required parameter. If folks wanna start top-level let's make it obvious they do so.

slashmo (approved):
LGTM

Sources/Tracing/Tracer.swift
ktoso: Nice catch, thank you!

And I'll fix the old swift issues.

Issue #67, opened 2022-10-06
startSpan and friends should take source location


ktoso:
This way we can build dynamically enable-able trace points.

PR #68, opened 2022-10-06
startSpan should take tracepoint location

ktoso:
resolves https://github.com/apple/swift-distributed-tracing/issues/67

This allows the implementation of ad-hoc tracerbullets and configurable per "trace location" tracers.

Here's the gist of it:

```swift
        tracer.enableTracepoint(fileID: fileID, line: line)
        // Imagine this is set via some "ops command", e.g. `<control> <pid, or ssh or something> trace enable Sample.swift:1234`
        // Effectively enabling tracepoints is similar to tracer bullets, tho bullets are generally "one off",
        // but here we could attach a trace-rate, so e.g.: control `<pid> trace enable Sample:1234 .2` to set 20% sampling rate etc.

        tracer.withSpan("dont") { span in
            // don't capture this span...
        }
        tracer.withSpan("yes", file: #fileID, line: 7777) { span in
            // do capture this span, and all child spans of it!
            tracer.withSpan("yes-inner") { inner in
                // since the parent of this span was captured, this shall be captured as well
            }
        }
```

or even better:

```
tracer.enableTracepoint(function: "example()")
```


ktoso:
Paging @weissi - the idea that we talked about yesterday. this would enable various location sensitive configuration or whatever one would want to do in a tracer

Sources/Tracing/Tracer.swift
```diff
@@ -30,11 +30,15 @@ public protocol Tracer: Instrument {
...
     ///   - kind: The ``SpanKind`` of the new ``Span``.
     ///   - time: The `DispatchTime` at which to start the new ``Span``.
+    ///   - file: The `fileID` where the span was started.
+    ///   - line: The file line where the span was started.
     func startSpan(
         _ operationName: String,
         baggage: Baggage,
         ofKind kind: SpanKind,
-        at time: DispatchWallTime
+        at time: DispatchWallTime,
```
weissi, line 39: why not the new Clock/Duration/... API?

ktoso: Yeah just noticed this too tbh...

I'll need to think about this, it'd force users to require 5.7+ so we may not want that... we'd want vapor to adopt without having to suddenly bump their required swift versions. Probably we can add an public API accepting `Instant` but I guess implementors may have to implement against dispatch time or something else... to keep the swift requirement low :-/ 

I'll think about it more and ticketify. AHC we'd also not want to force to require 5.7 suddenly 🤔

ktoso:

Sources/Tracing/Tracer.swift
ktoso: We can add an overload though that delegates to the dispatch time one 👍

Sources/Tracing/Tracer.swift
```diff
@@ -47,25 +53,111 @@ public protocol Tracer: Instrument {
 }
 
 extension Tracer {
-    /// Start a new ``Span`` with the given `Baggage` starting at `DispatchWallTime.now()`.
+    #if swift(>=5.3.0)
+    /// Start a new ``Span`` with the given `Baggage` starting "now".
```
slashmo, line 57: Good simplification 👍

Sources/Tracing/Tracer.swift
```diff
@@ -47,25 +53,111 @@ public protocol Tracer: Instrument {
...
     ///
     /// - Parameters:
     ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
     ///   - baggage: Baggage potentially containing trace identifiers of a parent ``Span``.
     ///   - kind: The ``SpanKind`` of the ``Span`` to be created. Defaults to ``SpanKind/internal``.
+    ///   - function: The function name in which the span was started
```
slashmo, line 63:
```suggestion
    ///   - function: The function name in which the span was started.
```

Sources/Tracing/Tracer.swift
```diff
@@ -47,25 +53,111 @@ public protocol Tracer: Instrument {
...
+    ///   - fileID: The `fileID` where the span was started.
+    ///   - line: The file line where the span was started.
     public func startSpan(
         _ operationName: String,
         baggage: Baggage,
-        ofKind kind: SpanKind = .internal
+        ofKind kind: SpanKind = .internal,
+        function: String = #function,
+        file fileID: String = #fileID,
```
slashmo, line 71: I'm very much looking forward to a time where we can start to only support `#fileID` in server-side Swift projects 😬

Sources/Tracing/Tracer.swift
ktoso: I'll add overload in separate PR

Sources/Tracing/Tracer.swift
```diff
@@ -47,25 +53,111 @@ public protocol Tracer: Instrument {
...
+    /// Execute a specific task within a newly created ``Span``.
+    ///
+    /// DO NOT `end()` the passed in span manually. It will be ended automatically when the `operation` returns.
+    ///
+    /// - Parameters:
+    ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
+    ///   - baggage: Baggage potentially containing trace identifiers of a parent ``Span``.
+    ///   - kind: The ``SpanKind`` of the ``Span`` to be created. Defaults to ``SpanKind/internal``.
+    ///   - operation: operation to wrap in a span start/end and execute immediately
+    ///   - function: The function name in which the span was started
```
slashmo, line 129:
```suggestion
    ///   - function: The function name in which the span was started.
```

Sources/Tracing/Tracer.swift
```diff
@@ -47,25 +53,111 @@ public protocol Tracer: Instrument {
...
+        )
     }
+    #else
+    /// Start a new ``Span`` with the given `Baggage` starting "now".
+    ///
+    /// - Parameters:
+    ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
+    ///   - baggage: Baggage potentially containing trace identifiers of a parent ``Span``.
+    ///   - kind: The ``SpanKind`` of the ``Span`` to be created. Defaults to ``SpanKind/internal``.
+    ///   - function: The function name in which the span was started
```
slashmo, line 91:
```suggestion
    ///   - function: The function name in which the span was started.
```

Sources/Tracing/Tracer.swift
```diff
@@ -75,15 +167,29 @@ extension Tracer {
     ///   - baggage: Baggage potentially containing trace identifiers of a parent ``Span``.
     ///   - kind: The ``SpanKind`` of the ``Span`` to be created. Defaults to ``SpanKind/internal``.
     ///   - operation: operation to wrap in a span start/end and execute immediately
+    ///   - function: The function name in which the span was started
```
slashmo, line 170:
```suggestion
    ///   - function: The function name in which the span was started.
```

Sources/Tracing/Tracer.swift
```diff
@@ -109,14 +216,27 @@ extension Tracer {
     ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
     ///   - kind: The ``SpanKind`` of the ``Span`` to be created. Defaults to ``SpanKind/internal``.
     ///   - operation: operation to wrap in a span start/end and execute immediately
+    ///   - function: The function name in which the span was started
```
slashmo, line 219:
```suggestion
    ///   - function: The function name in which the span was started.
```

Sources/Tracing/Tracer.swift
```diff
@@ -132,14 +252,27 @@ extension Tracer {
...
```
slashmo, line 255:
```suggestion
    ///   - function: The function name in which the span was started.
```

Sources/Tracing/Tracer.swift
```diff
@@ -162,15 +295,21 @@ extension Tracer {
     //               task local and modified before passing into this function. The baggage will be made the current task-local baggage for the duration of the `operation`.
     ///   - kind: The `SpanKind` of the `Span` to be created. Defaults to `.internal`.
     ///   - operation: operation to wrap in a span start/end and execute immediately
+    ///   - function: The function name in which the span was started
```
slashmo, line 298:
```suggestion
    ///   - function: The function name in which the span was started.
```

slashmo (approved):
LGTM

Sources/Tracing/Tracer.swift
ktoso: Tbh NIO requires 5.5+ already, so most projects can. I'm not even sure we should keep trying with 5.1+ here... but for now we do.

Sources/Tracing/Tracer.swift
ktoso: https://github.com/apple/swift-distributed-tracing/issues/71

PR #69, opened 2022-11-03
Documentation with latest guidance, and adjust types accordingly

ktoso:
Work in progress 

---

Revamp the docs and do the final renames before heading towards 1.0

Resolves https://github.com/apple/swift-distributed-tracing/issues/80


slashmo:
@ktoso Thanks for getting this started 🙏 I really like having the different DocC guides for the three different implementation use cases 👍

ktoso:
Yup, we'll separate it out and give lots of examples.

I'll spend time tomorrow on this and will ask for a review.

ktoso:
Today I rewrote the `Instrument Your Library or Framework` please have a look 👍

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,186 @@
...
+## Do you need an Instrument or a Tracer?
+
+Distributed tracing offers two types of instrumentation protocols: an instrument, and a tracer.
+
+A tracer is-an instrument as well, and further refines it with the ability to start a trace ``Span``.
+
+## Creating an instrument
+
+Creating an instrument means adopting the ``Instrument`` protocol (or ``Tracer`` in case you develop a tracer).
+`Instrument` is part of the `Instrumentation` library & `Tracing` contains the ``Tracer`` protocol.
```
yim-lee, line 18:
```suggestion
``InstrumentProtocol`` is part of the `Instrumentation` library and `Tracing` contains the ``TracerProtocol`` protocol.
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,186 @@
...
+
+`Instrument` has two requirements:
+
+1. A method to inject values inside a `FIXME!!!` into a generic carrier (e.g. HTTP headers)
```
yim-lee, line 22:
```suggestion
1. A method to inject values into a `FIXME!!!` a generic carrier (e.g. HTTP headers)
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,186 @@
...
+  InstrumentationSystem.instrument.inject(
+    context.baggage,
+    into: &request.headers,
+    using: HTTPHeadersInjector()
+  )
+}
+```
+
+On the receiving side, an HTTP server should use the following `Instrument` API to extract the HTTP headers of the given
+`HTTPRequest` into:
```
yim-lee, line 84:
```suggestion
On the receiving side, an HTTP server should use the following `Instrument` API to extract the HTTP headers of the given
`HTTPRequest`:
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,186 @@
...
+    request.headers,
+    into: &context.baggage,
+    using: HTTPHeadersExtractor()
+  )
+  // ...
+}
+```
+
+> In case your library makes use of the `NIOHTTP1.HTTPHeaders` type we already have an `HTTPHeadersInjector` &
+`HTTPHeadersExtractor` available as part of the `NIOInstrumentation` library.
```
yim-lee, line 98:
```suggestion
> In case your library makes use of the `NIOHTTP1.HTTPHeaders` type we already have an `HTTPHeadersInjector` and
`HTTPHeadersExtractor` available as part of the `NIOInstrumentation` library.
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,186 @@
...
+  self.execute(request).always { _ in
+    // set some more attributes & potentially record an error
+
+    // end the span
+    span.end()
+  }
+}
+```
+
+> ⚠️ Make sure to ALWAYS end spans. Ensure that all paths taken by the code will result in ending the span.
```
yim-lee, line 132:
```suggestion
> ⚠️ Make sure to ALWAYS end spans to ensure that all paths taken by the code will result in ending the span.
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,186 @@
...
+> Make sure that error cases also set the error attribute and end the span.
+
+> In the above example we used the semantic `http.method` attribute that gets exposed via the
+`TracingOpenTelemetrySupport` library.
+
+## Instrument developers: Creating an instrument
```
yim-lee, line 138: This is repeated info? L15

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
+# Instrument Your Library or Framework
+
+## Overview 
+
+This guide is aimed at library and framework developers who wish to instrument their code using distributed tracing.
+
+Doing so within a library may enable automatic trace propagation and is key to propagating trace information across
+distributed nodes, e.g. by instrumenting the HTTP client used by such system.
+
+Other examples of libraries which benefit _the most_ from being instrumented using distributed tracing include:
```
yim-lee, line 10:
```suggestion
Other examples of libraries which would benefit _the most_ from being instrumented using distributed tracing include:
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+- RPC systems (Swift gRPC or Swift's `DistributedActorSystem` implementations),
+- database drivers (e.g. SQL or MongoDB clients),
+- any other library which can emit meaningful span information about tasks it is performing.
+
+The most important libraries to instrument are "edge" libraries, which serve to connect between systems, because
+it is them who must inject and extract contextual baggage metadata to enable distributed trace ``Span`` propagation.
+
+Following those, any database or other complex library which may be able to emit useful information about its internals are
+also good candidates to being instrumented. Note that libraries may do so optionally, or hide the "verboseness" of such traces
+behind options, or only attach information if a Span is already active etc. Please review your library's documentation to learn
```
yim-lee, line 23:
```suggestion
behind options, or only attach information if a ``Span`` is already active etc. Please review your library's documentation to learn
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+more about it has integrated tracing support.
+
+### Propagating baggage metadata
+
+When crossing boundaries between processes, such as making or receiving an HTTP request, the library responsible for doing so, should invoke instrumentation in order to inject or extract the contextual baggage metadata into/from the "carrier" type (such as the `HTTPResponse`) type.
```
yim-lee, line 28:
```suggestion
When crossing boundaries between processes, such as making or receiving an HTTP request, the library responsible for doing so should invoke instrumentation in order to inject or extract the contextual baggage metadata into/from the "carrier" type (such as the `HTTPResponse`) type.
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+
+#### Handling outbound requests
+
+When a library makes an "outgoing" request or message interaction, it should invoke the  method of a configured instrument. This will invoke whichever instrument the end-user has configured and allow them to customize what metadata gets to be propagated. This can be by the following diagram:
```
yim-lee, line 32:
```suggestion
When a library makes an "outgoing" request or message interaction, it should invoke the method of a configured instrument. This will invoke whichever instrument the end-user has configured and allow them to customize what metadata gets to be propagated. This can be depicted by the following diagram:
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+  
+  try await _send(request)
+}
+```
+
+As you can see, the library does not know anything about what tracing system or instrumentation is installed, because it cannot know that ahead of time. 
+
+All it has to do is query for the current [task-local](https://developer.apple.com/documentation/swift/tasklocal) `Baggage` value, and if one is present, call on the instrumentation system to inject it into the request.
+
+Since neither, the tracing API, or the specific tracer backend are aware of this library's specific `HTTPRequest` type, we also need to implement an ``Injector`` which takes on the responsibility of adding the metadata into the carrier type (which in our case is the `HTTPRequest`). An injector could for example be implemented like this:
```
yim-lee, line 77:
```suggestion
Since neither the tracing API nor the specific tracer backend is aware of this library's specific `HTTPRequest` type, we also need to implement an ``Injector`` which takes on the responsibility of adding the metadata into the carrier type (which in our case is the `HTTPRequest`). An injector could for example be implemented like this:
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+
+```swift
+struct HTTPRequestInjector: Injector {
+    func inject(_ value: String, forKey key: String, into request: inout HTTPRequest) {
+        request.headers.append((key, value))
+    }
+}
+```
+
+Once the metadat has been injected, the request–including all the additional metadata–is sent over the network.
```
yim-lee, line 87:
```suggestion
Once the metadata has been injected, the request–including all the additional metadata is sent over the network.
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+
+> Note: The actual logic of deciding what baggage values to inject depend on the tracer implementation, and thus we are not covering it in this _end-user_ focused guide. Refer to <doc:ImplementATracer> if you'd like to learn about implementing a ``TracerProtocol``.
+
+#### Handling inbound requests
+
+On the receiving side, an HTTP server needs to perform the inverse operation, as it receives the request from the network and forms an `HTTPRequest` object, and before it passed it to user-code, it must extract any trace metadata from the request headers into the `Baggage`.
```
yim-lee, line 93:
```suggestion
On the receiving side, an HTTP server needs to perform the inverse operation, as it receives the request from the network and forms an `HTTPRequest` object, and before it passes it to user-code, it must extract any trace metadata from the request headers into the `Baggage`:
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+│ t  │          │┌───────────────┐  ┌─▼───────────────┐   │              
+│ w  │          ││     parse     │  │extract metadata │   │     ┌───────────┐
+│ o  │─────────▶││  HTTPRequest  ├──▶from HTTPRequest │───┼────▶│ User code │
+│ r  │          │└───────────────┘  └─────────────────┘   │     └───────────┘
+│ k  │          │                                         │              
+│    │          └─────────────────────────────────────────┘              
+└────
+```
+
+This is very similar to what we were doing on the outbound side, but the roles of baggage and request are somewhat inversed: we're extracting values from the carrier into the baggage. The code performing this task could look something like this:
```
yim-lee, line 116:
```suggestion
This is very similar to what we were doing on the outbound side, but the roles of baggage and request are somewhat reversed: we're extracting values from the carrier into the baggage. The code performing this task could look something like this:
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+    request,
+    into: &baggage,
+    using: HTTPRequestExtractor()
+  )
+
+  // ... invoke user code ...
+}
+```
+
+Similarily to the outbound side, we need to implement an ``Extractor`` because the tracing libraries don't know about our specific HTTP types, yet we need to have them decide for which values to extract keys.
```
yim-lee, line 135:
```suggestion
Similarly to the outbound side, we need to implement an ``Extractor`` because the tracing libraries don't know about our specific HTTP types, yet we need to have them decide for which values to extract keys.
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+  InstrumentationSystem.instrument.extract(..., into: &baggage, ...)
+  // ... 
+
+  await Baggage.withValue(baggage) {
+    await userCode(request)
+  }
+}
+```
+
+This sets the task-local value `Baggage.current` which is used by [swift-log](https://github.com/apple/swift-log), as well as ``TracerProtocol`` APIs in order to later "*pick up*" the baggage and and e.g. include it in log statements, or start new trace spans using the information stored in the baggage.
```
yim-lee, line 161:
```suggestion
This sets the task-local value `Baggage.current` which is used by [swift-log](https://github.com/apple/swift-log), as well as ``TracerProtocol`` APIs in order to later "*pick up*" the baggage and e.g. include it in log statements, or start new trace spans using the information stored in the baggage.
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+  
+  func onBodyPart(part: ByteBuffer) async { 
+    await Baggage.withValue(self.baggage) { 
+      await userHandler.requestHeaders(headers)
+    }
+  }
+}
+```
+
+While this code is very simple for illustration purposes, and it may seem suprising why there are those two separate places where we need to call into user-code separately. But in practice such situations can happen when using asynchronous network or database libraries which offer their API in terms of such callbacks. Always consider if and when to restore baggage such that it makes sense for the end user.
```
yim-lee, line 203:
```suggestion
While this code is very simple for illustration purposes, and it may seem surprising why there are two separate places where we need to call into user-code separately, in practice such situations can happen when using asynchronous network or database libraries which offer their API in terms of callbacks. Always consider if and when to restore baggage such that it makes sense for the end user.
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Starting Trace Spans in Your Library
+
+The above steps are enough if you wanted to provide contextual baggage propagation, and already enables techniques such as *correlation ids* which can be set once, in one system, and then carried through to any downstream services the code makes calls from while the baggage is set.
+
+Many libraries also have the opportunity to start trace spans themselfes, on behalf of users, in pieces of the library that can provide useful insight in the behavior or the library in production. For example, the HTTPServer can start spans as soon as it begins handling HTTP requests, and this way provide a parent span to any spans the user-code would be creating itself.
```
yim-lee, line 209:
```suggestion
Many libraries also have the opportunity to start trace spans themselves, on behalf of users, in pieces of the library that can provide useful insight in the behavior or the library in production. For example, the HTTPServer can start spans as soon as it begins handling HTTP requests, and this way provide a parent span to any spans the user-code would be creating itself. 
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Starting Trace Spans in Your Library
+
+The above steps are enough if you wanted to provide contextual baggage propagation, and already enables techniques such as *correlation ids* which can be set once, in one system, and then carried through to any downstream services the code makes calls from while the baggage is set.
+
+Many libraries also have the opportunity to start trace spans themselfes, on behalf of users, in pieces of the library that can provide useful insight in the behavior or the library in production. For example, the HTTPServer can start spans as soon as it begins handling HTTP requests, and this way provide a parent span to any spans the user-code would be creating itself. 
+
+Let us revisit the previous sample HTTPServer which restored baggage around invoking the user-code, and further extend it to start a span including basic information about the HTTPRequest being handled:
```
yim-lee, line 211:
```suggestion
Let us revisit the previous sample HTTPServer which restores baggage around invoking the user-code, and further extend it to start a span including basic information about the HTTPRequest being handled:
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,294 @@
...
+  userCode.handle(request)
+}
+
+func finishHandling(request: HTTPRequest, response: HTTPResponse) {
+  // always end spans (!)
+  span?.end() // ends and flushes the span
+}
+```
+
+It is very important to _always_ end spans that are started, as otherwise resources they use may keep accumulating and lead to memory leaks, or worse issues in tracing backends dependin on their implementation.
```
yim-lee, line 275:
```suggestion
It is very important to _always_ end spans that have been started, as otherwise resources they use may keep accumulating and lead to memory leaks, or worse issues in tracing backends depending on their implementation.
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,186 @@
...
+> In case your library makes use of the `NIOHTTP1.HTTPHeaders` type we already have an `HTTPHeadersInjector` &
+`HTTPHeadersExtractor` available as part of the `NIOInstrumentation` library.
+
+For your library/framework to be able to carry `FIXME!!!` across asynchronous boundaries, it's crucial that you carry the context throughout your entire call chain in order to avoid dropping metadata.
+
+### Tracing your library
+
+When your library/framework can benefit from tracing, you should make use of it by integrating the `Tracing` library.
+
+In order to work with the tracer [configured by the end-user](#Bootstrapping-the-Instrumentation-System), it adds a property to `InstrumentationSystem` that gives you back a `Tracer`. You can then use that tracer to start `Span`s. In an HTTP client you e.g.
```
yim-lee, line 106: Inconsistent use of end-user vs "end user"

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -0,0 +1,228 @@
...
+
+`FIXME!!!` should always be passed around explicitly.
+
+Libraries which support tracing are expected to accept a `FIXME!!!` parameter, which can be passed through the entire application. Make sure to always pass along the context that's previously handed to you. E.g., when making an HTTP request using `AsyncHTTPClient` in a `NIO` handler, you can use the `ChannelHandlerContext`s `baggage` property to access the `FIXME!!!`.
+
+#### Context argument naming/positioning
+
+> 💡 This general style recommendation has been ironed out together with the Swift standard library, core team, the SSWG as well as members of the community. Please respect these recommendations when designing APIs such that all APIs are able to "feel the same" yielding a great user experience for our end users ❤️
+>
+> It is possible that the ongoing Swift Concurrency efforts, and "Task Local" values will resolve this explicit context passing problem, however until these arrive in the language, please adopt the "context is the last parameter" style as outlined here.
```
yim-lee, line 95:
```suggestion
> It is possible that the ongoing Swift Concurrency efforts and "Task Local" values will resolve this explicit context passing problem, however until these arrive in the language, please adopt the "context is the last parameter" style as outlined here.
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
ktoso: Thanks, I didn't yet rewrite other guides other than the "Instrument your library" FYI

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
ktoso: Hmm, thanks I will reflow this entire guide 👍

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
ktoso: Or even Silmarillion-y 😂 Man, this word is my arch-enemy... thanks for spotting it

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -0,0 +1,443 @@
...
+// .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: 1.0.0),
+```
+
+Next, add the dependency to your application target:
+
+```swift
+.target(
+    name: "MyApplication",
+    dependencies: [
+      .product(name: "OpenTelemetry", package: "opentelemetry-swift"),
```
freef4ll, line 40: Maybe should include `OtlpGRPCSpanExporting` from very start?

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -0,0 +1,443 @@
...
+4. Finally, any other parts of your application
+
+This is because tracing systems may attempt to emit logs or metrics about their status etc.
+
+If you intend to use trace identifiers for log correlation (i.e. logging a `trace-id` in every log statement that is part of a trace),
+then don't forget to also configure a swift-lot `MetadataProvider`.
+
+A typical bootstrap could look something like this:
+
+```swift
```
freef4ll, line 102: Maybe keep this as a full example (with MultiThreadedEventLoopGroup)  and remove the bootstrapping the tracer above since this is a more complete example and having the two very close to each other is duplication?

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -0,0 +1,443 @@
...
+```
+
+`MultiplexInstrument` will then call out to each instrument it has been initialized with.
+
+### Introducing Trace Spans
+
+The primary way you interact with distributed tracing is by starting ``Span``s.
+
+Spans form hierarchies with their parent spans, and end up being visualized using various tools, usually in a format similar to gant charts. So for example, if we had multiple operations that compose making dinner, they would be modelled as child spans of a main `makeDinner` span. Any sub tasks are again modelled as child spans of any given operation, and so on.
+
```
freef4ll, line 154: Add link to https://opentelemetry.io/docs/concepts/observability-primer/ ?

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -0,0 +1,443 @@
...
+    public struct NestedSpanAttributes: NestedSpanAttributesProtocol {
+        public init() {}
+
+        /// HTTP request method. E.g. "GET".
+        public var method: Key<String> { "http.method" }
+    }
+}
+```
+
+### Span Events
```
freef4ll, line 400: Maybe move this section before `Span Attributes` section?

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
freef4ll: And also add a note about `BatchSpanProcessor` vs `SimpleSpanProcessor`. As simple processor is not suitable for hot code execution as will lead to 700+ byte message being sent each time the span is executed.

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
ktoso: Kind of through events are less "common" than attributes, and we do events + logging together, I'll keep as is.

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
ktoso: Yeha might as well, We're not "limited to" otel but they have some good guides, I'll add a links

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
ktoso: I'm a bit torn about this... This isn't the otel tracer's guide. For what it's worth this should be "go and read a specific tracer impls docs" but I want something good enough here to get someone started.

I'll see how to improve this bit though

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -0,0 +1,443 @@
...
+
+If you intend to use trace identifiers for log correlation (i.e. logging a `trace-id` in every log statement that is part of a trace),
+then don't forget to also configure a swift-lot `MetadataProvider`.
+
+A typical bootstrap could look something like this:
+
+```swift
+import OpenTelemetry
+import StatsdMetrics
+import FancyLogging
```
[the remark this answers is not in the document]
ktoso: Good catch, thanks

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
ktoso: For what it's worth this might better be "SomeTracingSystem" and not the otel not to confuse people. We just happen to use otel in the examples below, but this just explains adding a dependency 🤔  I'll think some more how to do this well.

ktoso:
I'm going to revive the docs now, but probably by making it a new PR

ktoso:
Replaced by https://github.com/apple/swift-distributed-tracing/pull/121

PR #72, opened 2022-12-02
Add .spi.yml for Swift Package Index DocC support

yim-lee:


.spi.yml
```diff
@@ -0,0 +1,4 @@
+version: 1
+builder:
+  configs:
+    - documentation_targets: [Tracing]
```
slashmo, line 4: I'm not very familiar with the Swift package index yet and its DocC support, but would this also surface the documentation from the `Instrumentation` target given that `Tracing` depends on it, or would we want to declare both as `documentation_targets` here?

yim-lee: I guess I am not sure either. Let me include it just in case.

tomerd (approved):

slashmo:
@yim-lee I just checked, with this SPI configuration when you click on "Documentation" on the SPI website it takes you to the `Instrumentation` documentation instead of `Tracing` without any UI to switch to the `Tracing` documentation. Changing the path works, but IMHO we should definitely optimize for folks seeing the `Tracing` documentation first. I don't know if it's enough to flip the order of the `documentation_targets` array to do so.

yim-lee:
@finestructure Please see [slashmo's comment above](https://github.com/apple/swift-distributed-tracing/pull/72#issuecomment-1337035952). Is there a way for us to specify the "landing" target?

finestructure:
That's down to the order - we're landing on the first target in the list IIRC!

finestructure:
I can manually trigger a doc re-build out of turn to verify without waiting for the 24 cooldown, just let me know.

yim-lee:
> That's down to the order - we're landing on the first target in the list IIRC!

Oh great! Thanks for the quick response @finestructure!

yim-lee:
@slashmo - opened https://github.com/apple/swift-distributed-tracing/pull/74

yim-lee:
@finestructure Can you please trigger a rebuild? Thanks in advance! 🙏

finestructure:
That's rebuilt with revision e8e428b6db9d2b1a4675feb5dbfbd947a7b5bc1c

finestructure:
Ok, both are there, got confused by our own UX there for a sec and couldn't find the target switcher 😅🤔

CleanShot 2022-12-05 at 19 29 16@2x

yim-lee:
Thanks @finestructure. You are awesome!

ktoso:
Great, thanks everyone!

Issue #79, opened 2023-01-27
Span should be marked as `Sendable`


fabianfett:
I'm pretty sure we want to allow `Span`s to be passed around. For this reason we should require them to be `Sendable`.

ktoso:
Agreed, will do

PR #83, opened 2023-01-30
Mark types Sendable

ktoso:
Resolves https://github.com/apple/swift-distributed-tracing/issues/79

cc @fabianfett

### DIFF
--- a/Sources/Instrumentation/Instrument.swift
+++ b/Sources/Instrumentation/Instrument.swift
@@ -24,7 +24,7 @@ public protocol _SwiftInstrumentationSendable {}
 /// Conforming types are used to extract values from a specific `Carrier`.
 public protocol Extractor: _SwiftInstrumentationSendable {
     /// The carrier to extract values from.
-    associatedtype Carrier: Sendable
+    associatedtype Carrier
 
     /// Extract the value for the given key from the `Carrier`.
     ///
@@ -37,7 +37,7 @@ public protocol Extractor: _SwiftInstrumentationSendable {
 /// Conforming types are used to inject values into a specific `Carrier`.
 public protocol Injector: _SwiftInstrumentationSendable {
     /// The carrier to inject values into.
-    associatedtype Carrier: Sendable
+    associatedtype Carrier
 
     /// Inject the given value for the given key into the given `Carrier`.
     ///
--- a/Sources/Instrumentation/Instrument.swift
+++ b/Sources/Instrumentation/Instrument.swift
@@ -14,10 +14,21 @@
 
 import InstrumentationBaggage
 
+/// Typealias used to simplify Support of old Swift versions which do not have `Sendable` defined.
+#if swift(>=5.6.0)
+@preconcurrency public protocol _SwiftInstrumentationSendable: Sendable {}
+#else
+public protocol _SwiftInstrumentationSendable {}
+#endif
+
 /// Conforming types are used to extract values from a specific `Carrier`.
-public protocol Extractor {
+public protocol Extractor: _SwiftInstrumentationSendable {
     /// The carrier to extract values from.
+    #if swift(>=5.6.0)
+    associatedtype Carrier: Sendable
+    #else
     associatedtype Carrier
+    #endif
 
     /// Extract the value for the given key from the `Carrier`.
     ///
@@ -28,9 +39,13 @@ public protocol Extractor {
 }
 
 /// Conforming types are used to inject values into a specific `Carrier`.
-public protocol Injector {
+public protocol Injector: _SwiftInstrumentationSendable {
     /// The carrier to inject values into.
+    #if swift(>=5.6.0)
+    associatedtype Carrier: Sendable
+    #else
     associatedtype Carrier
+    #endif
 
     /// Inject the given value for the given key into the given `Carrier`.
     ///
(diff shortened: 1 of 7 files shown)


Sources/Instrumentation/Instrument.swift
```diff
@@ -14,10 +14,21 @@
...
+#else
+public protocol _SwiftInstrumentationSendable {}
+#endif
+
 /// Conforming types are used to extract values from a specific `Carrier`.
-public protocol Extractor {
+public protocol Extractor: _SwiftInstrumentationSendable {
     /// The carrier to extract values from.
+    #if swift(>=5.6.0)
+    associatedtype Carrier: Sendable
```
slashmo, line 28: Can't we use the `_SwiftInstrumentationSendable` protocol as the associated type constraint?

Sources/Tracing/Span.swift
```diff
@@ -650,10 +650,26 @@ public struct SpanLink {
...
+
+#if compiler(>=5.6)
+@preconcurrency public protocol _SwiftTracingSendableSpan: Sendable {}
+#else
+public protocol _SwiftTracingSendableSpan {}
+#endif
+
+#if compiler(>=5.6)
+extension SpanAttributes: Sendable {}
+extension SpanAttribute: @unchecked Sendable {}
```
slashmo, line 669: Maybe also explain why `@unchecked Sendable`?

slashmo (approved):

Sources/Instrumentation/Instrument.swift
ktoso: That would force people to conform to `_SwiftInstrumentationSendable`, but we just need to force them to conform to `Sendable` 💡

Sources/Tracing/Span.swift
ktoso: will do

Sources/Instrumentation/Instrument.swift
slashmo: Aah, makes sense 👍

Sources/Instrumentation/Instrument.swift
```diff
@@ -28,9 +39,13 @@ public protocol Extractor {
...
 
 /// Conforming types are used to inject values into a specific `Carrier`.
-public protocol Injector {
+public protocol Injector: _SwiftInstrumentationSendable {
     /// The carrier to inject values into.
+    #if swift(>=5.6.0)
+    associatedtype Carrier: Sendable
+    #else
     associatedtype Carrier
+    #endif
```
fabianfett, line 48: Why has Carrier to be Sendable?

Sources/Tracing/Span.swift
fabianfett: Agree with Moritz here, for each `@unchecked` we should have explanation why it can be unchecked.

Sources/Tracing/Span.swift
```diff
@@ -650,10 +650,26 @@ public struct SpanLink {
...
+extension SpanStatus: Sendable {}
+extension SpanEvent: @unchecked Sendable {} // unchecked because of DispatchWallTime
```
fabianfett, line 671: Is DispatchWallTime used for locking?

fabianfett (commented):
Thanks for pushing this. One bigger question and some nits.

Sources/Instrumentation/Instrument.swift
ktoso: I guess it doesn't have to -- I'll remove

Sources/Tracing/Span.swift
ktoso: No, for the "wall time" time type. Swift's Clocks are missing functionality _and_ we would have to require 5.7 so sadly it's a no go either way... Though I'm still looking into using Swift's clock types

fabianfett: but why is it "unchecked because of `DispatchWallTime`"? Is `DispatchWallTime` not marked as Sendable? Shouldn't we use `@preconcurrency` in that case?

ktoso: Right, I just moved to that in [`e3c5bc1` (#83)](https://github.com/apple/swift-distributed-tracing/pull/83/commits/e3c5bc16ad151b54f7c218782f7ab2ac9186362f) :)

ktoso:
Nailed it now, I think... Last look @fabianfett ? :)

Tests/TracingTests/DynamicTracepointTracerTests.swift
```diff
@@ -288,3 +290,8 @@ extension DynamicTracepointTestTracer {
         }
     }
 }
+
+#if compiler(>=5.6.0)
+extension DynamicTracepointTestTracer: @unchecked Sendable {} // only intended for single threaded testing {}
+extension DynamicTracepointTestTracer.TracepointSpan: @unchecked Sendable {} // only intended for single threaded testing {}
```
stevapple, line 296: I suspect the `{}` suffix here is unnecessary?

Tests/TracingTests/TracedLockTests.swift
```diff
@@ -158,3 +159,8 @@ private final class TracedLockPrintlnTracer: Tracer {
         }
     }
 }
+
+#if compiler(>=5.6.0)
+extension TracedLockPrintlnTracer: Sendable {}
+extension TracedLockPrintlnTracer.TracedLockPrintlnSpan: @unchecked Sendable {} // only intended for single threaded testing {}
```
stevapple, line 165: Same for `{}` here.

ktoso: Thanks, cleaned up!

Issue #78, opened 2023-01-27
Consider `Span` to be an `associatedType` on `Tracer`


fabianfett:
Currently the `Tracer` protocol has the `startSpan` requirement. The return of `startSpan` is a `Span` which is a protocol. In other words we return an existential here. Instead we might want to consider making the Span an associated type and returning the concrete type:

```swift
public protocol Tracer: Instrument {
    associatedType Span: SpanProtocol // the current Span protocol maybe needs a rename to SpanProtocol for this?

    func startSpan(
        _ operationName: String,
        baggage: Baggage,
        ofKind kind: SpanKind,
        at time: DispatchWallTime,
        function: String,
        file fileID: String,
        line: UInt
    ) -> Span

    func forceFlush()
}
```

I can see no reasons, why a Tracer would return Spans of different types. If that was a requirement adopters could work around this, by using internal enums.

### Result

We preserve more type information if we work with a concrete Tracer.

slashmo:
Sounds good to me 👍 The rename to `SpanProtocol` also sounds good and actually fits nicely with @ktoso's rename of `Tracer` and `Instrument` to `*Protocol`: #69

ktoso:
Yeah, sounds like a good idea -- it does seem like we'll indeed adopt more `...Protocol` types, so this sounds good as well.

ktoso:
Done, I'll do one more rename though -- we'll keep `protocol Span`

Issue #65, opened 2022-10-03
Change Tracer to TracerProtocol and ad Tracer.withSpan {}


ktoso:
It is a bit too annoying to use `InstrumentationSystem.tracer` even though "correctly named".

Let's rename to make it pleasant to use and discover tracing.

ktoso:
Done in https://github.com/apple/swift-distributed-tracing/pull/93

PR #84, opened 2023-02-01
Rename protocols, use associatedtype Span and add Tracer.current

ktoso:
Resolves https://github.com/apple/swift-distributed-tracing/issues/78 by making `Span` an `associatedtype` on `TracerProtocol`. Since this caused entering this "...Protocol" naming, I did the same with Tracer and Instrument which we wanted to do anyway (see below).

Partially resolves: https://github.com/apple/swift-distributed-tracing/issues/65 though I did not add `Tracer.withSpan` yet -- that would bring back an existential there for the `any SpanProtocol` hm... It would allow writing `Tracer.withSpan() { ... }` rather than `Tracer.current.withSpan(){}` though -- do we think we'd like to add that, or it's just un-necessary repeating of the same exact APIs. For now keeping just `Tracer.current`.

I removed the outdated README since it caused more harm than good right now and updated the docc page a little bit. The full docs rewrite will land as: https://github.com/apple/swift-distributed-tracing/pull/69


ktoso:
Heh actually... this runs into the usual hell of:

```
protocol 'TracerProtocol' can only be used as a generic constraint because it has Self or associated type requirements
```

in old Swift versions... so that'll be very hard to support keeping source-compatibility between the language versions I think 🤔 Not sure we can pull this off nicely for `Span` 🤔

PR #85, opened 2023-02-02
[Discussion] prototype converting span attributes into metadata

ktoso:
@fabianfett noticed he ended up with logs containing:

`<date> info what : http_flavor=string("1.1") http_method=string("GET") http_status_code=string("405 Method Not Allowed") http_target=string("/") [What] Some message`

which wrongly have those `string()` in there.

The code used to produce this was this:

```swift
     var metadata = Logger.Metadata()

        if let httpSpan = trace.httpSpan {
            if let method = httpSpan.attributes["http.method"] {
                metadata["http_method"] = "\(method.toSpanAttribute())"
            }
            if let route = httpSpan.attributes["http.route"] {
                metadata["http_route"] = "\(route)"
            }
       }
```

So this PR is to discuss what we should do about this.

### DIFF
--- a/Sources/Tracing/Span.swift
+++ b/Sources/Tracing/Span.swift
@@ -269,7 +269,7 @@ public enum SpanAttribute: Equatable {
         .int(0)
     }
 
-    internal var anyValue: Any {
+    public var anyValue: Any {
         switch self {
         case .int(let value):
             return value
--- a/Tests/TracingTests/SpanTests.swift
+++ b/Tests/TracingTests/SpanTests.swift
@@ -197,6 +197,103 @@ final class SpanTests: XCTestCase {
         }
         XCTAssertEqual(statusCode, 418)
     }
+
+    func testAttributesAsStringMetadata() {
+        var attributes = SpanAttributes()
+        attributes["http.status_code"] = 418
+        attributes["http.target"] = "target"
+
+        var metadata: [String: SwiftLogMetadataValue] = [:]
+        metadata[attributeKey: "http_target"] = attributes["http.target"]
+        #if swift(>=5.2)
+        metadata[attributeKey: "http_status_code"] = attributes.sampleHttp.statusCode
+        #else
+        metadata[attributeKey: "http_status_code"] = attributes["http.status_code"]
+        #endif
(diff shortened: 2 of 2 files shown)


Sources/Tracing/Span.swift
```diff
@@ -269,7 +269,7 @@ public enum SpanAttribute: Equatable {
         .int(0)
     }
 
-    internal var anyValue: Any {
+    public var anyValue: Any {
```
ktoso, line 272: One way to get around this would be to make this property public and then the conversions would have to be...

```
            if let route = httpSpan.attributes["http.route"] {
                metadata["http_route"] = "\(route.anyValue)"
            }
```

This is pretty meh though; so continue reading

Tests/TracingTests/SpanTests.swift
```diff
@@ -197,6 +197,103 @@ final class SpanTests: XCTestCase {
...
+
+    func testAttributesAsStringMetadata() {
+        var attributes = SpanAttributes()
+        attributes["http.status_code"] = 418
+        attributes["http.target"] = "target"
+
+        var metadata: [String: SwiftLogMetadataValue] = [:]
+        metadata[attributeKey: "http_target"] = attributes["http.target"]
+        #if swift(>=5.2)
+        metadata[attributeKey: "http_status_code"] = attributes.sampleHttp.statusCode
```
ktoso, line 209: This reads much better:

```swift
        metadata[attributeKey: "http_status_code"] = attributes.sampleHttp.statusCode
```

but we're back in the problem of one library having to depend on the other in order to provide the conversions. When we hit this earlier we thought that tracing depending on logging would not be too bad since it is the "heavier" of the two, so we could provide the conversions here...

Arguably the subscript is problematic here though, because of the differences in expressive power -- so we can't easily roundtrip (See the impls below). So maybe this should be:

```swift
metadata.setAttribute(attributes.sampleHttp.statusCode)
metadata.setAttribute("http_status_code", attributes.sampleHttp.statusCode)
```

?

ktoso:
Backtracking on this -- I think we can do this better after all...

Issue #71, opened 2022-11-07
Adopt Clock in tracing APIs


ktoso:

ktoso:
Turns out this is pretty tricky, I'll look into it some more

ktoso:
We may be able to hide away the Dispatch dependency here by introducing a `TracingTime` type that would have only a "now" and unix timestamp initializer. If we get an UTCClock this is easier to add support for it, and deprecate the entire type, rather than pass around the Dispatch types explicitly.

Thanks @weissi for the idea.

ktoso:
This seems to be turning into "abstract over Clocks" as we're ending up in problematic spaces otherwise.

PR #86, opened 2023-02-02
Abstract away time source: TracingTime

ktoso:
Introduce a `TracingTime` type that hides the dispatch use for the time source.
If we get a Clock that can do this, we can more easily move to using it in API, without breaking tracer implementations.

Kind of... Resolves https://github.com/apple/swift-distributed-tracing/issues/71

### DIFF
--- a/Sources/Tracing/Span.swift
+++ b/Sources/Tracing/Span.swift
@@ -12,11 +12,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if swift(>=5.6.0)
-@preconcurrency import struct Dispatch.DispatchWallTime
-#else
-import struct Dispatch.DispatchWallTime
-#endif
 @_exported import InstrumentationBaggage
 
 /// A `Span` represents an interval from the start of an operation to its end, along with additional metadata included
@@ -67,7 +62,7 @@ public protocol Span: AnyObject, _SwiftTracingSendableSpan {
     /// - Parameter time: The `DispatchWallTime` at which the span ended.
     ///
     /// - SeeAlso: `Span.end()` which automatically uses the "current" time.
-    func end(at time: DispatchWallTime)
+    func end(at time: TracingTime)
 }
 
 extension Span {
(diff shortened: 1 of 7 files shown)


Sources/Tracing/Span.swift
```diff
@@ -108,15 +103,15 @@ public struct SpanEvent: Equatable {
     /// One or more ``SpanAttribute``s with the same restrictions as defined for ``Span`` attributes.
     public var attributes: SpanAttributes
 
-    /// The `DispatchWallTime` at which this event occurred.
-    public let time: DispatchWallTime
+    /// The time at which this event occurred.
+    public let time: TracingTime
```
ktoso, line 107: The important part here is that even if we deprecate Span creation with this type and move to Clock, it would not have to be breaking to Tracer implementations -- they can still work without rushing to a breaking change in time type

Sources/Tracing/TracingTime.swift
```diff
@@ -0,0 +1,72 @@
...
+/// standard library (or elsewhere) that we might want to use as source of time.
+///
+/// Abstracting away the source of the measurement into this wrapper allows us to
+/// have ``Tracer`` implementations don't care about the user API that is surfaced as `startSpan`
+/// to end-users. As time goes on, this API may evolve, but we will not have to make breaking changes
+/// to tracer implementations themselves.
+public struct TracingTime: Sendable {
+  enum Repr {
+      case dispatchWallTime(DispatchWallTime)
+  }
```
ktoso, line 35: potentially other Clock.Instants or other time sources here...

Sources/Tracing/TracingTime.swift
```diff
@@ -0,0 +1,72 @@
...
+  private var repr: Repr
+
+  private init(dispatchWallTime: DispatchWallTime) {
```
weissi, line 38: why not public?

ktoso: i attempted to hide away dispatch completely, debatable if we care to do so or not

ktoso:
replaced by https://github.com/apple/swift-distributed-tracing/pull/98

Issue #48, opened 2021-04-13
Allow mutation of span operation name


slashmo:
We might want to consider offering an API to rename an already started span. This is [part of the OTel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#updatename) but comes with a disadvantage for sampling.

### Use-Case: HTTP Server

HTTP servers might want to rename a span after creation, because they can come up with a more relevant span in the meantime.

1. Immediately start span when receiving a request (`operationName` = `request.head.uri`)
2. Route request
  2.1. Route Not Found --> Record error
  2.2 Route Found --> Rename to route (`/users/1` becomes `/users/:userID`)
3. End span

This approach has one disadvantage. In a typical tracer with head-first sampling, sampling decisions based on the `operationName` of a span will potentially not be accurate as the sampler won't be invoked after the rename.

### Alternatives: Start spans after the fact

As mentioned in [OTel Spec: `Span.UpdateName`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#updatename) an alternative to renaming an already started span would be to record the timestamp whenever the span should have been started, and then passing that timestamp to the `startSpan` call.

1. Get current timestamp
2. Route request
  2.1. Route Not Found --> Start span at previous timestamp (`operationName` = `request.head.uri`) --> Record error
  2.2 Route Found --> Start span at previous timestamp (`operationName` = `/users/:userID`)
3. End span

PR #88, opened 2023-02-24
Allow for reading and mutating a span name

ktoso:
**Motivation:**

The feature frozen otel specification allows for renaming spans after they have been started. This often is too late to impact sampling decisions, however we still can use to to provide better names for spans as time goes on.

**Modifications:**

add an operationName to the Span protocol.

A span has to be thread-safe already and we have mutating operations on it already so it does not change the complexity requirements on implementing a span.

**Result:**

Be closer to otel spec and more flexible in naming and renaming spans.
Resolves https://github.com/apple/swift-distributed-tracing/issues/48

### DIFF
--- a/Sources/Tracing/NoOpTracer.swift
+++ b/Sources/Tracing/NoOpTracer.swift
@@ -50,7 +50,7 @@ public struct NoOpTracer: Tracer {
         public let baggage: Baggage
         public let isRecording = false
 
-        public let _operationName: String
+        private let _operationName: String
         public var operationName: String {
             get {
                 self._operationName
--- a/Sources/Tracing/NoOpTracer.swift
+++ b/Sources/Tracing/NoOpTracer.swift
@@ -29,7 +29,7 @@ public struct NoOpTracer: Tracer {
         file fileID: String,
         line: UInt
     ) -> Span {
-        NoOpSpan(baggage: baggage)
+        NoOpSpan(operationName: operationName, baggage: baggage)
     }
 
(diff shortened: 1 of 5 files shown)


ktoso:
If unsure about this feature, please check the https://github.com/apple/swift-distributed-tracing/issues/48 discussion and otel's justification which I think is reasonable https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#updatename

slashmo (approved):
LGTM

Sources/Tracing/NoOpTracer.swift
```diff
@@ -50,7 +50,18 @@ public struct NoOpTracer: Tracer {
         public let baggage: Baggage
         public let isRecording = false
 
-        public init(baggage: Baggage) {
+        public let _operationName: String
+        public var operationName: String {
```
stevapple, line 54: Should this be `public`?

ktoso: Good catch, it should not :)

Sources/Tracing/NoOpTracer.swift
```diff
@@ -50,7 +50,18 @@ public struct NoOpTracer: Tracer {
         public let baggage: Baggage
         public let isRecording = false
 
-        public init(baggage: Baggage) {
+        public let _operationName: String
```
ktoso, line 53:
```suggestion
        private let _operationName: String
```

PR #90, opened 2023-02-24
Recording an error may carry attributes

ktoso:
**Motivation:**
It can be used to attach more information bout the error, like its stacktrace and similar information.

This is also supported by the otel spec, so we're aligning closer with it here.

**Modifications:**

Change the recordError to accept optional span attributes

**Result:**

more powerful record error API

resolves https://github.com/apple/swift-distributed-tracing/issues/45

### DIFF
--- a/Sources/Tracing/Span.swift
+++ b/Sources/Tracing/Span.swift
@@ -42,7 +42,7 @@ public protocol Span: AnyObject, _SwiftTracingSendableSpan {
     ///
     /// - Parameters:
     ///   - error: The error to be recorded.
-    func recordError(_ error: Error)
+    func recordError(_ error: Error, attributes: SpanAttributes)
 
     /// The attributes describing this `Span`.
     var attributes: SpanAttributes { get set }
@@ -94,6 +94,12 @@ extension Span {
     /// - Parameter attributes: The ``SpanAttributes`` describing this link. Defaults to no attributes.
     public func addLink(_ other: Span, attributes: SpanAttributes = [:]) {
         self.addLink(SpanLink(baggage: other.baggage, attributes: attributes))
+    }
+}
+
+extension Span {
+    public func recordError(_ error: Error) {
+        self.recordError(error, attributes: [:])
     }
 }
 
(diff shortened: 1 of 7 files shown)


Sources/Tracing/Span.swift
```diff
@@ -42,7 +42,7 @@ public protocol Span: AnyObject, _SwiftTracingSendableSpan {
     ///
     /// - Parameters:
     ///   - error: The error to be recorded.
```
slashmo, line 44:
```suggestion
    /// - Parameters:
    ///   - error: The error to be recorded.
    ///   - attributes: Additional attributes describing the error.
```

Sources/Tracing/Span.swift
```diff
@@ -97,6 +97,12 @@ extension Span {
     }
 }
 
+extension Span {
+    public func recordError(_ error: Error) {
```
slashmo, line 101:
```suggestion
    /// Record a failure described by the given error.
    ///
    /// - Parameters:
    ///   - error: The error to be recorded.
    public func recordError(_ error: Error) {
```

Tests/TracingTests/TracerTests.swift
```diff
@@ -253,6 +253,37 @@ final class TracerTests: XCTestCase {
...
+        tracer.withSpan("hello") { span in
+            span.recordError(errorToThrow, attributes: attrsForError)
+        }
+
+        XCTAssertTrue(endedSpan != nil)
+        XCTAssertEqual(endedSpan!.recordedErrors.count, 1)
+        let error = endedSpan!.recordedErrors.first!.0
+        XCTAssertEqual(error as! ExampleSpanError, errorToThrow)
+        let attrs = endedSpan!.recordedErrors.first!.1
+        XCTAssertEqual(attrs, attrsForError)
```
slashmo, line 283: Do we want to use `XCTUnwrap` here instead of the force unwraps?

slashmo (approved):
Thanks for tackling this 👍

Tests/TracingTests/TracerTests.swift
ktoso: Meh, tbh I don't mind those `!`, thank you for the review!

PR #91, opened 2023-02-24
Incomplete idea how to solve the Clock trouble

ktoso:
This aims to showcase one way to solve: https://github.com/apple/swift-distributed-tracing/issues/71

We'd have to drop anything older than Swift 5.7, but we'd be able to express things over a generic Clock and use the Tracing clock for now as an impl detail

### DIFF
--- a/Sources/Tracing/Span.swift
+++ b/Sources/Tracing/Span.swift
@@ -26,7 +26,10 @@ import struct Dispatch.DispatchWallTime
 /// Creating a `Span` is delegated to a ``Tracer`` and end users should never create them directly.
 ///
 /// - SeeAlso: For more details refer to the [OpenTelemetry Specification: Span](https://github.com/open-telemetry/opentelemetry-specification/blob/v0.7.0/specification/trace/api.md#span) which this type is compatible with.
-public protocol Span: AnyObject, _SwiftTracingSendableSpan {
+@available(macOS 13.0, iOS 15.0, *)
+public protocol SpanProtocol<Clock>: AnyObject, _SwiftTracingSendableSpan {
+    associatedtype Clock: _Concurrency.Clock where Clock.Duration == Swift.Duration
+
     /// The read-only `Baggage` of this `Span`, set when starting this `Span`.
     var baggage: Baggage { get }
 
--- a/Sources/Tracing/Span.swift
+++ b/Sources/Tracing/Span.swift
@@ -70,7 +70,7 @@ public protocol SpanProtocol<Clock>: AnyObject, _SwiftTracingSendableSpan {
     /// - Parameter time: The `DispatchWallTime` at which the span ended.
     ///
     /// - SeeAlso: `Span.end()` which automatically uses the "current" time.
-    func end(at time: DispatchWallTime)
+    func end(at time: Clock.Instant)
 }
 
 extension Span {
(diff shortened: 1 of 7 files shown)


Sources/Tracing/Span.swift
```diff
@@ -26,7 +26,10 @@ import struct Dispatch.DispatchWallTime
 /// Creating a `Span` is delegated to a ``Tracer`` and end users should never create them directly.
 ///
 /// - SeeAlso: For more details refer to the [OpenTelemetry Specification: Span](https://github.com/open-telemetry/opentelemetry-specification/blob/v0.7.0/specification/trace/api.md#span) which this type is compatible with.
-public protocol Span: AnyObject, _SwiftTracingSendableSpan {
+@available(macOS 13.0, iOS 15.0, *)
+public protocol SpanProtocol<Clock>: AnyObject, _SwiftTracingSendableSpan {
+    associatedtype Clock: _Concurrency.Clock where Clock.Duration == Swift.Duration
```
ktoso, line 31: "The change"™

Sources/Tracing/Tracer.swift
```diff
@@ -41,7 +44,7 @@ public protocol Tracer: Instrument {
         function: String,
         file fileID: String,
         line: UInt
-    ) -> Span
+    ) -> Self.Span
```
ktoso, line 47: We'll be returning concrete spans then, as requested in another issue

Sources/Tracing/TracingClock.swift
```diff
@@ -0,0 +1,49 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@available(macOS 13.0, iOS 15.0, *)
+public struct TracingClock: _Concurrency.Clock {
```
ktoso, line 16: We have to implement a TracingClock that gives us unit timestamps for "now", since there is no clock implementation in Swift which does this.

If we get one, we'd deprecate this type immediately

Sources/Tracing/TracingClock.swift
```diff
@@ -0,0 +1,49 @@
...
+    public var now: Instant {
+        .init(raw: 0) // TODO: implement "now"
+    }
+    public var minimumResolution: Duration {
+        .milliseconds(1)
+    }
+
+    public func sleep(until deadline: Instant, tolerance: Duration?) async throws {
+        fatalError("Not implemented for TracingClock")
+    }
```
ktoso, line 47: Potentially, we could not implement sleeping on it at all tbh

Sources/Tracing/Span.swift
```diff
@@ -67,7 +70,7 @@ public protocol Span: AnyObject, _SwiftTracingSendableSpan {
     /// - Parameter time: The `DispatchWallTime` at which the span ended.
     ///
     /// - SeeAlso: `Span.end()` which automatically uses the "current" time.
-    func end(at time: DispatchWallTime)
+    func end(at time: Clock.Instant)
```
ktoso, line 73: And this way we got rid of the Dispatch types in API - which would have stayed with us "forever"

stevapple (commented):
I would have loved this change if it didn’t require Swift 5.7+…

It is elegant and really solves the problem. Swift 5.7 requirement is not too bad for server applications since NIO only preserves ~1year support for older Swift versions, but it may affect the adoption on client side (especially with iOS 15 requirement).

PR #92, opened 2023-03-13
associated type Span, Tracer as short-hand, and *Protocol types

ktoso:
**Motivation:**

This is a revival of
https://github.com/apple/swift-distributed-tracing/pull/84 where we try to KEEP compatibility with versions below 5.7 with a compatibility "legacy" tracer type, but otherwise move towards requiring 5.7 for all the "nice" apis that use associated types and `any
  TracerProtocol` and friends

**Modifications:**

- `Tracer` -> `TracerProtocol`
- `Tracer` is now a namespace in order to `Tracer.withSpan {}` easily
- `Span` -> `SpanProtocol`
- Introduce `LegacyTracerProtocol` which does not make use of associated type Span, and can be used in 5.6 libraries; they can deprecate and move away form it ASAP as they start requiring 5.7

**Result:**

Offer the APIs we want in 5.7 but remain compatible with 5.6 until we drop it as soon as 5.9 is released as stable - this allows us to adopt eagerly in libraries without having to wait for 5.9 to drop.

Replaces https://github.com/apple/swift-distributed-tracing/pull/84

### DIFF
--- a/Sources/Tracing/SpanProtocol.swift
+++ b/Sources/Tracing/SpanProtocol.swift
@@ -99,7 +99,6 @@ public protocol SpanProtocol: AnyObject, _SwiftTracingSendableSpan {
     /// - SeeAlso: `Span.end()` which automatically uses the "current" time.
     // @available(*, deprecated, message: "Use Clock based `end(at:)` instead")
     func end(at time: DispatchWallTime)
-
 }
 
 extension SpanProtocol {
--- a/Sources/Tracing/SpanProtocol.swift
+++ b/Sources/Tracing/SpanProtocol.swift
@@ -25,8 +25,17 @@ import struct Dispatch.DispatchWallTime
 ///
 /// Creating a `Span` is delegated to a ``Tracer`` and end users should never create them directly.
 ///
+/// ### Reference semantics
+/// A span always must exhibit reference semantics. Passing around a `span` must allow other pieces of code
+/// modify it safely. The span must therefore employ synchronization techniques adequate to ensure this.
+///
+/// It is allowed to implement `SpanProtocol` using a `struct` or `enum`, however the type must still exhibit reference
+/// semantics. This is useful especially for implementing efficient "no-op" or "tracing is disabled" span implementations,
+/// which can be almost empty struct instances, while their "tracing is enabled" versions should refer all state to an
+/// underlying reference type "box" which contains all the spans data.
+///
 /// - SeeAlso: For more details refer to the [OpenTelemetry Specification: Span](https://github.com/open-telemetry/opentelemetry-specification/blob/v0.7.0/specification/trace/api.md#span) which this type is compatible with.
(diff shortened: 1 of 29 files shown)


Sources/Tracing/SpanProtocol.swift
```diff
@@ -26,7 +26,7 @@ import struct Dispatch.DispatchWallTime
 /// Creating a `Span` is delegated to a ``Tracer`` and end users should never create them directly.
 ///
 /// - SeeAlso: For more details refer to the [OpenTelemetry Specification: Span](https://github.com/open-telemetry/opentelemetry-specification/blob/v0.7.0/specification/trace/api.md#span) which this type is compatible with.
-public protocol Span: AnyObject, _SwiftTracingSendableSpan {
```
fabianfett, line 29: Something that has just come to my mind, if we try to make not using a tracer as noop as possible we should lift the `AnyObject` requirement here. This way we can implement Noops with empty spans.

ktoso: Not sure that makes sense... Semantically they really are reference types.

You also wanted to return _concrete types_ making it not possible to return `NoopSpan` "sometimes" and the "real one" usually. since that's different types.

So I guess the only trick this could pull then is to have a struct or enum that is "the noop one or the real one that contains a class `box` with the actual impl. I'm assuming we'd keep the methods not mutating even then, since value semantics are _wrong_ for spans.

fabianfett: > Not sure that makes sense... Semantically they really are reference types.

100% agree. In NIO we sometimes use structs that wrap reference types and we document heavily that they have reference semantics. 

> to have a struct or enum that is "the noop one or the real one that contains a class box with the actual impl. I'm assuming we'd keep the methods not mutating even then, since value semantics are wrong for spans.

Yes! 100% this.

ktoso: This "let's pretend `nonmutating`" won't work, I feel like I've been through this multiple times already if I'm honest.

I won't do this dance in this PR.

Things break e.g. here:

```swift
        let tracer: any TracerProtocol = InstrumentationSystem.tracer

        var baggage = Baggage.topLevel
        InstrumentationSystem.instrument.extract(request.headers, into: &baggage, using: HTTPHeadersExtractor())

        let span = tracer.startSpan("GET \(request.path)", baggage: baggage)

        let response = self.catchAllHandler(span.baggage, request, self.client)
        span.attributes["http.status"] = response.status
```

ktoso:
Can't merge this since we stopped running 5.5 CI but this PR requires it 😆 

Making a new PR

Sources/Tracing/SpanProtocol.swift
ktoso: solved in https://github.com/apple/swift-distributed-tracing/pull/94

Issue #82, opened 2023-01-30
SpanAttributes subscript should use concrete type and not an existential


fabianfett:
Currently `SpanAttributes` uses an existential for its subscript API. 

```swift
public struct SpanAttributes {
  public subscript(_ name: String) -> SpanAttributeConvertible? { get set }
}
```

Swift Logging uses the concrete type `Logger.Metadata.Value` instead:
```swift
public struct Logger {
    subscript(metadataKey metadataKey: String) -> Logger.MetadataValue? { get set }

    public enum MetadataValue {
        case string(String)
        case stringConvertible(CustomStringConvertible & Sendable)
        case dictionary(Metadata)
        case array([Metadata.Value])
    }
}
```

I think that we should change the SpanAttributes subscript API to return a `SpanAttribute`. This way this would work like Logging and we remove an existential from the API boundary

ktoso:
Thanks for catching this -- here I think indeed we'll be able to do that 👀 Let's give it a shot.

ktoso:
I'm not sure about this, it means we'd lose the ability to:

```
        span.attributes["http.status"] = response.status

```

which is very nice 🤔 I could not find a way around it.

ktoso:
We can add additional API though, I'll do that

PR #95, opened 2023-03-20
Provide explicit set/get to avoid SpanAttributeConvertible wrapping

ktoso:
Provide explicit set/get to avoid SpanAttributeConvertible existential wrapping, to those who care about such overhead

**Motivation:**

Provide more control and allow avoiding the existential SpanAtrributeConvertible access when using the subscript.

**Modifications:**

additional API to avoid the existential

**Result:**
There is a way to use the API completely avoiding existentials

resolves https://github.com/apple/swift-distributed-tracing/issues/82

### DIFF
--- a/Sources/Tracing/SpanProtocol.swift
+++ b/Sources/Tracing/SpanProtocol.swift
@@ -537,6 +537,16 @@ extension SpanAttributes {
         }
     }
 
+    /// Similar to `subscript(_:)` however returns the stored `SpanAttribute` rather than going through `SpanAttributeConvertible`.
+    public func get(_ name: String) -> SpanAttribute? {
+        self._attributes[name]
+    }
+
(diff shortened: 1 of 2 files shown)


Tests/TracingTests/SpanTests+XCTest.swift
yim-lee, line 1: I think we can do without this file now that this project is Swift 5.6+?

yim-lee (approved):

Tests/TracingTests/SpanTests+XCTest.swift
ktoso: I was going to check :-) Yes I think so, right? I'll remove

ktoso: https://github.com/apple/swift-distributed-tracing/pull/97

PR #97, opened 2023-03-20
remove linux test shims

ktoso:
**Motivation:**
no longer needed since we require 5.6+

**Modifications:**
remove the linux tests generator

**Result:**

cleaner build


yim-lee (approved):
🎉

PR #98, opened 2023-03-23
Implement TracerClock and abstract away "now"

ktoso:
**Motivation:**
we do not want to be bound to DispatchWallTime, but we cannot require Clock. We need to introduce a shim type TracerClockProtocol that in the future we'll be able to conform a "good Swift clock" to.

This removes any dependencies on dispatch and makes the Time source swappable. We also implement our own unix timestamp clock.

The core of the change is:
Screenshot 2023-03-23 at 22 53 37

resolves https://github.com/apple/swift-distributed-tracing/issues/71

WIll get to clean this up and make all versions work soon. At least locally confirmed swiftlang-5.7.0.123.8 worked fine


ktoso:
Just need to get 5.6 done in the overloads maddness and we're done here 🥳

Sources/Tracing/SpanProtocol.swift
```diff
@@ -154,17 +148,28 @@ public struct SpanEvent: Equatable {
     public var attributes: SpanAttributes
 
     /// The `DispatchWallTime` at which this event occurred.
-    public let time: DispatchWallTime
+    public let time: UInt64
```
FranzBusch, line 151: Can we add some docs what this timestamp is e.g. millis since Epoch? Interestingly enough should we make this opaque or not?

Sources/Tracing/Tracer.swift
```diff
@@ -152,55 +235,76 @@ extension Tracer {
...
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) // for TaskLocal Baggage
+    public static func withSpan<T, Clock: TracerClock>(
+        _ operationName: String,
+        clock: Clock,
+        baggage: @autoclosure () -> Baggage = .current ?? .topLevel,
+        ofKind kind: SpanKind = .internal,
+        function: String = #function,
+        file fileID: String = #fileID,
+        line: UInt = #line,
+        _ operation: (any Span) async throws -> T
```
FranzBusch, line 254: Aren't we missing a `@_inheritActorContext` and `@Sendable` annotation here. Otherwise this could runs into warnings right?


```
actor Foo {
  var bar = 0 
  func foo() async {
    await withSpan {
      self.bar += 1
    }
  }
}
```

The downside of adding these annotations is that the story inside classes/structs is becoming worse because they don't have any actor context so the `@Sendable` constraint makes it impossible for them to access state even though the `operation` closure is run "inline".

In the end, we really need a productionized version of `@_unsafeInheritExecutor`. We are running into the same problem over in service-lifecycle.

Sources/Tracing/TracingTime.swift
```diff
@@ -0,0 +1,186 @@
...
+}
+
+extension SwiftDistributedTracingDurationProtocol {
+    public static func /= (_ lhs: inout Self, _ rhs: Int) {
+        lhs = lhs / rhs
+    }
+}
+
+public protocol SwiftDistributedTracingInstantProtocol: Comparable, Hashable, Sendable {
+    // associatedtype Duration: SwiftDistributedTracingDurationProtocol
```
FranzBusch, line 40: Can we remove this?

Sources/Tracing/TracingTime.swift
```diff
@@ -0,0 +1,186 @@
...
+}
+
+public protocol TracerInstantProtocol: SwiftDistributedTracingInstantProtocol {
+    /// Representation of this instant as the number of milliseconds since UNIX Epoch (January 1st 1970)
+    @inlinable
+    var millisSinceEpoch: UInt64 { get }
```
FranzBusch, line 46: Do we want to spell this out?

```suggestion
    var millisecondsSinceEpoch: UInt64 { get }
```

FranzBusch (commented):
Some small comments

Sources/Tracing/TracingTime.swift
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    public var now: Self.Instant {
+        var ts = timespec()
+        clock_gettime(CLOCK_REALTIME, &ts)
+        /// We use unsafe arithmetic here because `UInt64.max` nanoseconds is more than 580 years,
+        /// and the odds that this code will still be running 530 years from now is very, very low,
+        /// so as a practical matter this will never overflow.
+        let nowNanos = UInt64(ts.tv_sec) &* 1_000_000_000 &+ UInt64(ts.tv_nsec)
+        let nowMillis = UInt64(nowNanos / 1_000_000) // nanos to millis
```
FranzBusch, line 107: I thought we were going to use a different method to calculate UTC time?

Sources/Tracing/Tracer.swift
ktoso: I checked and we don't get warnings for this 🤔

Sources/Tracing/TracingTime.swift
ktoso: Ok :)

Sources/Tracing/TracingTime.swift
ktoso: Yeah I had a whole complicated dance which... ended up at the same number ¯\_(ツ)_/¯ 

Discussed with @Lukasa and we're unsure why Foundation does the extreme dance there; This is good enough for us. And we could support nanos even but for now I didn't

Sources/Tracing/SpanProtocol.swift
ktoso: Opaque is rather tricky what it would mean... Practically speaking we need timestamps, a tracer could decide it needs some specific way of those but by the end of the day it all ends up as ints I guess.

Making it all a type bound to `Tracer` is a huge mess type wise if we need to retain the 5.6 support :-\

ktoso:
Thanks for the review @FranzBusch I'm going to merge so we can give this a round of testing

PR #101, opened 2023-03-29
Default LegacyTracerProtocol implementation

fabianfett:
### Motivation

For adopters that implement the `TracerProtocol` we should offer a default implementation for the `LegacyTracerProtocol`, so that they don't need to worry about it at all.

### Changes

- Add default `LegacyTracerProtocol` implementation on `TracerProtocol`


ktoso:

Issue #110, opened 2023-03-30
Rename TracerProtocol.TracerSpan -> TracerProtocol.Span


fabianfett:
We like shorter names.

PR #111, opened 2023-03-30
consider TracerSpan assoc type -> Span

ktoso:
Something @fabianfett requested we give another look at.

Resolves https://github.com/apple/swift-distributed-tracing/issues/110 but we should discuss this more -- it comes with a confusing tradeoff for implementers of spans I suppose. Please discuss. I'm on the fence, it is nice to not have `Tracer.TracerSpan` but the tradeoff is some potential confusion hmmm

**Motivation:**

Users who use generic `Tracer` end up having to spell `Tracer.TracerSpan`.

We could help those situations by renaming the associated type. However at the cost of causing name conflicts where the Span implementation must qualify that it is implementing `Tracing.Span`.

This may or may not be worth it -- discuss.

**Modifications:**

`associatedtype TracerSpan: Span` -> `associatedtype Span: Tracing.Span`

### DIFF
--- a/Sources/Tracing/NoOpTracer.swift
+++ b/Sources/Tracing/NoOpTracer.swift
@@ -19,7 +19,7 @@ import Dispatch
 /// Tracer that ignores all operations, used when no tracing is required.
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) // for TaskLocal Baggage
 public struct NoOpTracer: LegacyTracer {
-    public typealias TracerSpan = NoOpSpan
+    public typealias Span = NoOpSpan
 
     public init() {}
 
@@ -29,7 +29,7 @@ public struct NoOpTracer: LegacyTracer {
                                                  clock: Clock,
                                                  function: String,
                                                  file fileID: String,
-                                                 line: UInt) -> any Span
+                                                 line: UInt) -> any Tracing.Span
     {
         NoOpSpan(baggage: baggage())
     }
@@ -48,7 +48,7 @@ public struct NoOpTracer: LegacyTracer {
         // no-op
     }
 
-    public struct NoOpSpan: Span {
+    public struct NoOpSpan: Tracing.Span {
         public let baggage: Baggage
         public var isRecording: Bool {
             false
--- a/Sources/Tracing/TracerProtocol+Legacy.swift
+++ b/Sources/Tracing/TracerProtocol+Legacy.swift
@@ -54,7 +54,7 @@ public protocol LegacyTracer: Instrument {
         function: String,
         file fileID: String,
         line: UInt
-    ) -> any Span
+    ) -> any Tracing.Span
 
     /// Export all ended spans to the configured backend that have not yet been exported.
     ///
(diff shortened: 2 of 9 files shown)


fabianfett:
@ktoso can you give an example for this?

> We could help those situations by renaming the associated type. However at the cost of causing name conflicts where the Span implementation must qualify that it is implementing Tracing.Span.

Sources/Tracing/Tracer.swift
```diff
@@ -59,7 +59,7 @@ extension Tracer {
         function: String = #function,
         file fileID: String = #fileID,
         line: UInt = #line
-    ) -> any Span {
```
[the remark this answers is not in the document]
ktoso: Yeah here I can likely just do Span, that's right.

In a `TracerProtocol` we sometimes had to be explicit.

I can give this a cleanup pass

Sources/Tracing/NoOpTracer.swift
```diff
@@ -48,7 +48,7 @@ public struct NoOpTracer: LegacyTracerProtocol {
         // no-op
     }
 
-    public struct NoOpSpan: Span {
+    public struct NoOpSpan: Tracing.Span {
```
ktoso, line 51: This is one annoying spot, people MUST write Tracing.Span here

stevapple: What about swapping `TracerSpan` and `Span`?

ktoso: Maybe, what do @tomerd @fabianfett @Lukasa think?

fabianfett: I don't find this to annoying. The tough part here is really only on the tracer implementers. TracerSpan is on all the users. If we want to keep the pain small, the first group should deal with it, not the second.

ktoso: Yeah it's on implementors to do the dance, I don't mind either I guess.

stevapple: I would prefer `TracerSpan` to `[Tracing.]Span` here. Swift’s name lookup is still tricky now (see https://forums.swift.org/t/pitch-fully-qualified-name-syntax/28482 for the context). If the client defines a top-level `Tracing` type, they will completely lose access to `Tracing.Span`. Not to say `TracerSpan` is shorter by two ASCII characters:)

ktoso:
Yeah I think this will be fine. I'll merge tomorrow probably tho

Sources/Tracing/NoOpTracer.swift
stevapple: A typical tracing user doesn’t use either `[Tracing.]Span` or `Tracer.Span`, as the `Tracer.Span` type can be inferred from `Tracer.startSpan`.

Both `Span` and `TracerSpan` can cause some confusion with `Tracer.Span`, but the fact that `Span` is ambiguous, despite being possible to work around and comfort the compiler, will inevitably troubles code review.

Given that the type name is not used very often, I still suggests the longer but less error-prone `TracerSpan`.

ktoso:
Let's give this a shot

Issue #103, opened 2023-03-29
Remove SwiftDistributedTracingDurationProtocol


ktoso:
We don't require duration at all actually, just instants

PR #114, opened 2023-03-31
Add nanoseconds to instant protocol

ktoso:
**Motivation:**

we have nanoseconds precision on most platforms so lets expose it.

resolves https://github.com/apple/swift-distributed-tracing/issues/105

### DIFF
--- a/Sources/Tracing/TracingTime.swift
+++ b/Sources/Tracing/TracingTime.swift
@@ -21,12 +21,12 @@ import Darwin
 @_exported import Instrumentation
 @_exported import InstrumentationBaggage
 
-public protocol TracerInstantProtocol: Comparable, Hashable, Sendable {
+public protocol TracerInstant: Comparable, Hashable, Sendable {
     /// Representation of this instant as the number of nanoseconds since UNIX Epoch (January 1st 1970)
     var nanosecondsSinceEpoch: UInt64 { get }
 }
 
-extension TracerInstantProtocol {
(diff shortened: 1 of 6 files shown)


ktoso:
Also resolves https://github.com/apple/swift-distributed-tracing/issues/103 by removing unused protocol

Sources/Tracing/TracingTime.swift
```diff
@@ -21,26 +21,17 @@ import Darwin
...
-    static func *= (_ lhs: inout Self, _ rhs: Int)
-
-    static func / (_ lhs: Self, _ rhs: Self) -> Double
-}
-
-extension SwiftDistributedTracingDurationProtocol {
-    public static func /= (_ lhs: inout Self, _ rhs: Int) {
-        lhs = lhs / rhs
-    }
+public protocol TracerInstantProtocol: Comparable, Hashable, Sendable {
```
stevapple, line 24: This is one of the two last places where we have the `Protocol` suffix. Do we want to get rid of it entirely?

ktoso: Yeah I think we can :+1:

I'll do this in this PR actually. Implementations tend to be called Timestamp or something like that after all

Lukasa (approved):

fabianfett (approved):

slashmo (approved):

ktoso:
Sweet, thanks folks; moving on then

Issue #119, opened 2023-04-12
`addEvent` and `recordError` should take `clock` argument


slashmo:
`addEvent` and its related `recordError` method should take a `clock` argument, defaulting to `DefaultTracerClock()`. This would allow us to mock the clock similarly to how we can already do it when calling `Span.end()`.

ktoso:
Agreed, on it :+1

ktoso:
in addEvent no, because the Event creation will have it:

```
    func addEvent(_ event: SpanEvent)
```
```
    public init<Instant: TracerInstant>(name: String,
                                    at instant: @autoclosure () -> Instant,
                                    attributes: SpanAttributes = [:])
    {
```

record error is a good call tho, added:

```

    func recordError<Instant: TracerInstant>(_ error: Error,
                                             attributes: SpanAttributes,
                                             at instant: @autoclosure () -> Instant)

    public func recordError(_ error: Error, attributes: SpanAttributes) {
        self.recordError(error, attributes: attributes, at: DefaultTracerClock.now)
    }
```

Issue #108, opened 2023-03-30
Drop TracerClock from start/withSpan APIs, but use internally


fabianfett:
```swift
public protocol TracerProtocol: LegacyTracerProtocol {
    /// The concrete type of span this tracer will be producing/
    associatedtype Span: Tracing.Span

    /// Start a new ``Span`` with the given `Baggage`.
    ///
    /// The current task-local `Baggage` is picked up and provided to the underlying tracer.
    /// It is also possible to pass a specific `baggage` explicitly, in which case attempting
    /// to pick up the task-local baggage is prevented. This can be useful when we know that
    /// we're about to start a top-level span, or if a span should be started from a different,
    /// stored away previously,
    ///
    /// - Note: Prefer ``withSpan(_:baggage:ofKind:at:function:file:line:operation:)`` to start
    ///   a span as it automatically takes care of ending the span, and recording errors when thrown.
    ///   Use `startSpan` iff you need to pass the span manually to a different
    ///   location in your source code to end it.
    ///
    /// - Warning: You must `end()` the span when it the measured operation has completed explicitly,
    ///   otherwise the span object will potentially never be released nor reported.
    ///
    /// - Parameters:
    ///   - operationName: The name of the operation being traced. This may be a handler function, database call, ...
    ///   - baggage: The `Baggage` providing information on where to start the new ``Span``.
    ///   - kind: The ``SpanKind`` of the new ``Span``.
    ///   - clock: The clock to use as time source for the start time of the ``Span``
    ///   - function: The function name in which the span was started
    ///   - fileID: The `fileID` where the span was started.
    ///   - line: The file line where the span was started.
    func startSpan(
        _ operationName: String,
        baggage: @autoclosure () -> Baggage,
        ofKind kind: SpanKind,
        clock: some TracerInstantProtocol,
        function: String,
        file fileID: String,
        line: UInt
    ) -> TracerSpan

    func startSpan(
        _ operationName: String,
        baggage: @autoclosure () -> Baggage,
        ofKind kind: SpanKind,
        function: String,
        file fileID: String,
        line: UInt
    ) -> TracerSpan
}

extension TracerProtocol {
    func startSpan(
        _ operationName: String,
        baggage: @autoclosure () -> Baggage,
        ofKind kind: SpanKind,
        function: String,
        file fileID: String,
        line: UInt
    ) -> TracerSpan {

    }
}
```

PR #120, opened 2023-04-13
remove clock protocol, use autoclosure "now" instant

ktoso:
**Motivation:**

we don't need the clock protocol, but we do need the default clock to provide an instant.

We also add time to recordError in this PR

Resolves https://github.com/apple/swift-distributed-tracing/issues/119
Resolves https://github.com/apple/swift-distributed-tracing/issues/108

### DIFF
--- a/Sources/Tracing/NoOpTracer.swift
+++ b/Sources/Tracing/NoOpTracer.swift
@@ -23,13 +23,13 @@ public struct NoOpTracer: LegacyTracer {
 
     public init() {}
 
-    public func startAnySpan<Clock: TracerClock>(_ operationName: String,
-                                                 baggage: @autoclosure () -> Baggage,
-                                                 ofKind kind: SpanKind,
-                                                 clock: Clock,
-                                                 function: String,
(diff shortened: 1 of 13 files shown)


ktoso:
We have discussed this one with @fabianfett yesterday at length, it'll work nicely

ktoso:
I'll merge but post-merge review is welcome -- this mostly just swaps `clock: some TracerClock` for `at instant: @autoclosure () -> some TracerInstant` but we default impl using TracerClock so we can still mock it out globally if we wanted to 👍

Sources/Tracing/NoOpTracer.swift
```diff
@@ -84,19 +84,19 @@ public struct NoOpTracer: LegacyTracer {
             }
         }
 
-        public func end<Clock: TracerClock>(clock: Clock) {
+        public func end<Instant: TracerInstant>(at instant: Instant) {
```
fabianfett, line 87: do we need an `@autoclosure` here as well?

Issue #80, opened 2023-01-27
`Span` references `LoggingContext` in its docs


fabianfett:
```swift
/// A `Span` represents an interval from the start of an operation to its end, along with additional metadata included
/// with it. A `Span` can be created from a `Baggage` or `LoggingContext` which MAY contain existing span identifiers,
/// in which case this span should be considered as "child" of the previous span.
```

We should remove the ref to `LoggingContext` here.

slashmo:
I'm pretty sire that's not the only reference to the "old" context propagation mechanism. Konrad already started updating the documentation, not sure though if this particular reference was already removed: https://github.com/apple/swift-distributed-tracing/pull/69

ktoso:
Yeah this will be solved by the docs rewrite, ongoing here: https://github.com/apple/swift-distributed-tracing/pull/69

Since you also have a few things we can fix right away perhaps the plan forward is to keep merging and releasing 0.x things until we're happy and you've proven it out in adoption a bit :-)

PR #121, opened 2023-04-13
[Docs] Combined initial documentation work

ktoso:
Replaces https://github.com/apple/swift-distributed-tracing/pull/69

Resolves https://github.com/apple/swift-distributed-tracing/issues/80

### DIFF
--- a/Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
+++ b/Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
@@ -16,7 +16,7 @@ A complete [list of swift-distributed-tracing implementations](http://github.com
 is available in this project's README. Select an implementation you'd like to use and follow its bootstrap steps.
 
 > Note: Since instrumenting an **application** in practice will always need to pull in an existing tracer implementation,
-> in this guide we'll use the community maintained [`otel-swift`](https://github.com/slashmo/otel-swift) 
+> in this guide we'll use the community maintained [`swift-otel`](https://github.com/slashmo/swift-otel) 
 > tracer, as an example of how you'd start using tracing in your real applications.
 > 
 > If you'd rather implement your own tracer, refer to <doc:ImplementATracer>.


Samples/Dinner/docker/collector-config.yaml
```diff
@@ -0,0 +1,24 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: otel-collector:4317
+
+exporters:
+  logging:
+    logLevel: debug
```
slashmo, line 9: Just a heads-up that this changed to `verbosity: detailed` in a recent `otel-collector` release. `logLevel: debug` is not supported anymore. Since we're using the `latest` tag in `docker-compose.yml` we should update it accordingly.

```suggestion
    verbosity: detailed
```

ktoso: Nice, thanks! my docker setup os broken on my main box so I've not verified those yet :(

Samples/Dinner/Sources/Onboarding/main.swift
```diff
@@ -0,0 +1,63 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift OpenTelemetry open source project
+//
+// Copyright (c) 2021 Moritz Lang and the Swift OpenTelemetry project authors
```
ktoso, line 5: Authorship I believe you'd be fine with the samples using this repo's copyright @slashmo ?

Docs/README.md
```diff
@@ -0,0 +1,3 @@
+# Documentation landing page
+
+This documentation target exists only to serve as a landing page for self hosted documentation.
```
yim-lee, line 3:
```suggestion
This documentation target exists only to serve as a landing page for self-hosted documentation.
```

Docs/SwiftDistributedTracing/Docs.docc/index.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Overview
+
+This is a collection of Swift libraries enabling the instrumentation of server side applications using tools such as tracers. Our goal is to provide a common foundation that allows to freely choose how to instrument systems with minimal changes to your actual code.
+
+While Swift Distributed Tracing allows building all kinds of _instruments_, which can co-exist in applications transparently, its primary use is instrumenting multi-threaded and distributed systems with Distributed Traces.
+
+### Quickstart Guides
+
+We provide a number of guides aimed at getting your started with tracing your systems and have prepared them from three "angles":
```
yim-lee, line 13:
```suggestion
We provide a number of guides aimed at getting you started with tracing your systems and have prepared them from three "angles":
```

Docs/SwiftDistributedTracing/empty.swift
```diff
@@ -0,0 +1,19 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Distributed Tracing open source project
+//
+// Copyright (c) 2020-2021 Apple Inc. and the Swift Distributed Tracing project
```
yim-lee, line 5:
```suggestion
// Copyright (c) 2023 Apple Inc. and the Swift Distributed Tracing project
```

Samples/Dinner/README.md
```diff
@@ -0,0 +1,18 @@
+## Tracing Dinner Sample App
+
+In order to try this sample app, and visualize traces it produces, you should first run docker-compose in order
+to launch a docker containers which host a Zipkin UI and collector:
```
yim-lee, line 4:
```suggestion
In order to try this sample app, and visualize the traces it produces, you should first run `docker-compose` to launch a docker container which hosts a Zipkin UI and collector:
```

Samples/Dinner/README.md
```diff
@@ -0,0 +1,18 @@
...
+docker-compose -f docker/docker-compose.yaml up --build
+```
+
+and then run the sample app which will produce a number of traces:
+
+```
+swift run -c release
+```
+
+Refer to the "Trace Your Application" guide in the documentation to learn more about how to interpret this sample.
```
yim-lee, line 18:
```suggestion
Refer to the "Trace Your Application" guide to learn more about how to interpret this sample.
```

Would it be possible to link "Trace Your Application"?

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,186 @@
...
+
+This guide is aimed at ``TracerProtocol`` and ``InstrumentProtocol`` implementation authors.
+
+This guide is for you if you find yourself in need of implementing your own tracing client such as Zipkin, Jaeger, X-Trace, OpenTelemetry or something similar that is custom to your company or distributed system.
+
+## Do you need an Instrument or a Tracer?
+
+Distributed tracing offers two types of instrumentation protocols: an instrument, and a tracer.
+
+A tracer is-an instrument as well, and further refines it with the ability to start a trace ``Span``.
```
yim-lee, line 13:
```suggestion
A tracer is an instrument as well, and further refines it with the ability to start a trace ``Span``.
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,186 @@
+# Implement a Tracer
+
+## Overview
+
+This guide is aimed at ``TracerProtocol`` and ``InstrumentProtocol`` implementation authors.
```
yim-lee, line 5: IIRC we renamed them to `Tracer` and `Instrument`?

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,186 @@
...
+    // end the span
+    span.end()
+  }
+}
+```
+
+> ⚠️ Make sure to ALWAYS end spans to ensure that all paths taken by the code will result in ending the span.
+> Make sure that error cases also set the error attribute and end the span.
+
+> In the above example we used the semantic `http.method` attribute that gets exposed via the
```
yim-lee, line 135:
```suggestion
> In the above example we use the semantic `http.method` attribute that gets exposed via the
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,186 @@
...
+`TracingOpenTelemetrySupport` library.
+
+## Instrument developers: Creating an instrument
+
+Creating an instrument means adopting the `Instrument` protocol (or `Tracer` in case you develop a tracer).
+`Instrument` is part of the `Instrumentation` library & `Tracing` contains the `Tracer` protocol.
```
yim-lee, line 141:
```suggestion
`Instrument` is part of the `Instrumentation` library and `Tracing` contains the `Tracer` protocol.
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,331 @@
...
+behind options, or only attach information if a ``Span`` is already active etc. Please review your library's documentation to learn
+more about it has integrated tracing support.
+
+### Propagating baggage metadata
+
+When crossing boundaries between processes, such as making or receiving an HTTP request, the library responsible for doing so should invoke instrumentation in order to inject or extract the contextual baggage metadata into/from the "carrier" type (such as the `HTTPResponse`) type.
+
+#### Handling outbound requests
+
+When a library makes an "outgoing" request or message interaction, it should invoke the method of a configured instrument. This will invoke whichever instrument the end-user has configured and allow them to customize what metadata gets to be propagated. This can be depicted by the following diagram:
```
yim-lee, line 32:
```suggestion
When a library makes an "outgoing" request or message interaction, it should invoke the method of a configured instrument. This will invoke whichever instrument the end-user has configured and allow them to customize what metadata gets to be propagated. This is depicted in the following diagram:
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,331 @@
...
+
+```swift
+struct HTTPRequestInjector: Injector {
+    func inject(_ value: String, forKey key: String, into request: inout HTTPRequest) {
+        request.headers.append((key, value))
+    }
+}
+```
+
+Once the metadata has been injected, the request–including all the additional metadata–is sent over the network.
```
yim-lee, line 87:
```suggestion
Once the metadata has been injected, the request--including all the additional metadata--is sent over the network.
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,331 @@
...
+
+```swift
+struct HTTPRequestExtract: Instrumentation.Extractor {
+    func extract(key: String, from request: HTTPRequest) -> String? {
+        request.headers.first(where: { $0.0 == key })?.1
+    }
+}
+```
+
+Which exact keys will be asked for depends on the tracer implementation, thus we don't present this part of the implementation in part of the guide. For example, a tracer most likely would look for, and extract, values for keys such as `trace-id` and `span-id`. Note though that the exact semantics and keys used by various tracers differ, which is why we have to leave the decision of what to extract up to tracer implementations.
```
yim-lee, line 145:
```suggestion
Which exact keys will be asked for depends on the tracer implementation, thus we don't present this part of the implementation in this guide. For example, a tracer most likely would look for, and extract, values for keys such as `trace-id` and `span-id`. Note though that the exact semantics and keys used by various tracers differ, which is why we have to leave the decision of what to extract up to tracer implementations.
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,331 @@
...
+
+  await Baggage.withValue(baggage) {
+    await userCode(request)
+  }
+}
+```
+
+This sets the task-local value `Baggage.current` which is used by [swift-log](https://github.com/apple/swift-log), as well as ``TracerProtocol`` APIs in order to later "*pick up*" the baggage and e.g. include it in log statements, or start new trace spans using the information stored in the baggage.
+
+> Note: The end goal here being that when end-users of your library write `log.info("Hello")` the logger is able to pick up the contextual baggage information and include the e.g. the `trace-id` in such log statement automatically! This way, every log made during the handling of this request would include the `trace-id` automatically, e.g. like this:
```
yim-lee, line 163:
```suggestion
> Note: The end goal here being that when end-users of your library write `log.info("Hello")` the logger is able to pick up the contextual baggage information and include e.g. the `trace-id` in such log statement automatically! This way, every log made during the handling of this request would include the `trace-id` automatically, e.g. like this: 
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,331 @@
...
+}
+```
+
+While this code is very simple for illustration purposes, and it may seem surprising why there are two separate places where we need to call into user-code separately, in practice such situations can happen when using asynchronous network or database libraries which offer their API in terms of callbacks. Always consider if and when to restore baggage such that it makes sense for the end user.
+
+### Starting Trace Spans in Your Library
+
+The above steps are enough if you wanted to provide contextual baggage propagation. It already enables techniques such as **correlation ids** which can be set once, in one system, and then carried through to any downstream services the code makes calls from while the baggage is set.
+
+Many libraries also have the opportunity to start trace spans themselfes, on behalf of users, in pieces of the library that can provide useful insight in the behavior or the library in production. For example, the `HTTPServer` can start spans as soon as it begins handling HTTP requests, and this way provide a parent span to any spans the user-code would be creating itself.
```
yim-lee, line 209:
```suggestion
Many libraries also have the opportunity to start trace spans themselves, on behalf of users, in pieces of the library that can provide useful insight in the behavior or the library in production. For example, the `HTTPServer` can start spans as soon as it begins handling HTTP requests, and this way provide a parent span to any spans the user-code would be creating itself. 
```

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -0,0 +1,439 @@
...
+
+```swift
+withSpan("Working on my novel") { span in 
+  write(.novel)
+}
+```
+
+This API is available both in synchronous and asynchronous contexts. The closure also is passed a `span` object which is a mutable, but `Sendable ` object that tracer implementations must provide. A `Span` is an in memory representation of the trace span that can be enriched with various information about this execution. For example, if the span represents an HTTP request, one would typically add **span attributes** for `http.method`, `http.path` etc. 
+
+Throwing out of the operation closure automatically records an error in the `span`
```
yim-lee, line 270:
```suggestion
Throwing out of the operation closure automatically records an error in the `span`.
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
ktoso: Yes sorry I should have been explicit that not all code is cleanedup here yet

Samples/Dinner/README.md
```diff
@@ -0,0 +1,18 @@
+## Tracing Dinner Sample App
+
+In order to try this sample app, and visualize traces it produces, you should first run docker-compose in order
```
MaxDesiatov, line 3: formatting nit

```suggestion
In order to try this sample app, and visualize traces it produces, you should first run `docker-compose` in order
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
MaxDesiatov: I'm seeing this still in the diff, maybe this comment shouldn't be resolved?

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,169 @@
...
+A tracer is-an instrument as well, and further refines it with the ability to start a trace ``Span``.
+
+## Creating an instrument
+
+In order to implement an instrument you need to implement the `Instrument` protocol.
+`Instrument` is part of the `Instrumentation` library that `Tracing` depends on and offers the minimal core APIs that allow implementing instrumentation types.
+
+`Instrument` has two requirements:
+
+1. A `Instrument/extract(_:into:using:)` method, which extracts values from a generic carrier (e.g. HTTP headers) and store them into a `Baggage` instance
```
MaxDesiatov, line 22:
```suggestion
1. An `Instrument/extract(_:into:using:)` method, which extracts values from a generic carrier (e.g. HTTP headers) and store them into a `Baggage` instance
```

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,169 @@
...
+2. A `Instrument/inject(_:into:using:)` method, which takes values from the `Baggage` to inject them into a generic carrier (e.g. HTTP headers)
```
MaxDesiatov, line 23:
```suggestion
2. An `Instrument/inject(_:into:using:)` method, which takes values from the `Baggage` to inject them into a generic carrier (e.g. HTTP headers)
```

Samples/Dinner/README.md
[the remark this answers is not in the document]
ktoso: Yeah okey

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -0,0 +1,345 @@
...
+withSpan("event-\(event.id)", baggage: span.baggage) { span in
+  // ... 
+}
+```
+
+which is equivalent to surrounding thr withSpan with a binding of the baggage. The passed baggage (with the values updated by the tracer), will then be set for the duration of the `withSpan` operation, just like usual.
+
+### Global vs. "Stored" Tracers and Instruments
+
+Tracing functions similarly to swift-log and swift-metrics, in the sense that there is a global "backend" configured at application start, by end-users (developers) of an application. And this is how using `InstrumentationSystem/tracer` gets the "right" tracer at runtime.
```
MaxDesiatov, line 341: missing word here?

```suggestion
Tracing functions work similarly to swift-log and swift-metrics, in the sense that there is a global "backend" configured at application start, by end-users (developers) of an application. And this is how using `InstrumentationSystem/tracer` gets the "right" tracer at runtime.
```

slashmo: The way I read it here is that "it functions" is a substitute for "it works" instead of referring to functions as in methods.

ktoso: Right yeap it's as a verb here.

MaxDesiatov: Maybe better to rephrase as "Tracing works" to avoid this potential ambiguity?

ktoso: Yeah that's fair -- will do 👍

porglezomp (commented):
I went through part of the application example and ran into the fact that it seems like the example is already the parallelized trace, despite what the docs suggest. Is that expected? It seems like it might be nice to have it be able to produce both kinds of traces so that someone following the tutorial can actually explore the comparison features, etc.

Samples/Dinner/Package.swift
```diff
@@ -0,0 +1,23 @@
+// swift-tools-version:5.3
+import PackageDescription
+
+let package = Package(
+    name: "onboarding",
+    platforms: [
+        .macOS("14.0.0"),
```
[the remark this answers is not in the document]
ktoso: I'll drop to 13 thanks

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -0,0 +1,505 @@
...
+> $ cd Samples/Dinner
+> $ docker-compose -f docker/docker-compose.yaml up --build
+> # Starting docker_zipkin_1 ... done
+> # Starting docker_jaeger_1 ... done
+> # Recreating docker_otel-collector_1 ... done
+> # Attaching to docker_jaeger_1, docker_zipkin_1, docker_otel-collector_1
+> ```
+>
+> - Zipkin UI URL: http://localhost:9411/ 
+> - Jaeger UI URL: ...
```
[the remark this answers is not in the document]
ktoso: Thanks I'll remove those `-` here, we mention it above right

ktoso:
I take the few minor comments as overall approval, thanks folks!

MaxDesiatov (approved):

Samples/Dinner/Sources/Onboarding/Dinner.swift
```diff
@@ -0,0 +1,90 @@
...
+    async let veggies = try chopVegetables()
+    async let meat = marinateMeat()
+    async let oven = preheatOven(temperature: 350)
+    // ...
+    return try await cook(veggies, meat, oven)
+  }
+}
+
+func chopVegetables() async throws -> [Vegetable] {
+  try await otelChopping1.tracer().withSpan("chopVegetables") { _ in
```
Mordil, line 43: I know I'm coming in late, but how come this is explicitly referencing `otelChopping1`?

Sources/Tracing/Docs.docc/Guides/ImplementATracer.md
```diff
@@ -0,0 +1,169 @@
...
+A tracer is-an instrument as well, and further refines it with the ability to start a trace ``Span``.
+
+## Creating an instrument
+
+In order to implement an instrument you need to implement the `Instrument` protocol.
+`Instrument` is part of the `Instrumentation` library that `Tracing` depends on and offers the minimal core APIs that allow implementing instrumentation types.
+
+`Instrument` has two requirements:
+
+1. An `Instrument/extract(_:into:using:)` method, which extracts values from a generic carrier (e.g. HTTP headers) and store them into a `Baggage` instance
```
Mordil, line 22: I'm assuming this is referenced this way, so that when cross-module symbols are able to be referenced by DocC, we can just come back and add the 2nd set of `?

ktoso: Yeah currently we can't cross link so it's just a code marker. I may go around and make them explicit links manually hmm

Samples/Dinner/Sources/Onboarding/Dinner.swift
ktoso: Just to mock a multi node trace in one process. I'll make an actual multi process example soon to complement this with more "normal" usage

Mordil: Oh, got it. That wasn't explicitly clear when looking at this example what it was actually demonstrating

scripts/docs/preview_docc.sh
```diff
@@ -27,4 +26,4 @@
 ##
 ##===----------------------------------------------------------------------===##
 
-swift package --disable-sandbox preview-documentation --target $1
+xcrun swift package --disable-sandbox preview-documentation --target Tracing
```
Mordil, line 29: Well this doesn't seem that portable 😛

MaxDesiatov: What would be the alternative? Is there a portable way to pick up the currently selected toolchain? I don't think non-Darwin platforms even allow something like that.

Mordil: Platform checks to determine if it should be prefixed with `xcrun`

MaxDesiatov: Potentially, I'm not sure if the intention was to create a script that runs purely on macOS CI.

PR #129, opened 2023-06-01
Rename do not switch case

fabianfett:


Sources/Tracing/SpanProtocol.swift
```diff
@@ -327,7 +327,7 @@ public enum SpanAttribute: Equatable {
     case string(String)
     case stringArray([String])
 
-    case __DO_NOT_SWITCH_EXHAUSTIVELY_OVER_THIS_ENUM
+    case __DO_NOT_SWITCH_EXHAUSTIVELY_OVER_THIS_ENUM_USE_DEFAULT_INSTEAD
```
FranzBusch, line 330: Maybe we can add some written comment to this case as well clarifying this a bit more?

FranzBusch (approved):
We can add the comment in a follow up

Sources/Tracing/SpanProtocol.swift
ktoso: We'll add in a follow up

ktoso: We'll add in a follow up

ktoso (approved):

PR #130, opened 2023-06-05
Remove platform requirement from Package.swift

adam-fowler:
This PR removes the platform requirement from the Package.swift and replaces it with `@available` checks throughout the library where they weren't already placed.
I placed everything in the Benchmarks package inside a `DSLBenchmarks` enum so I only needed to add one `@available`.

These changes are required if Hummingbird is going to merge in its tracing middleware PR without making it a breaking change


.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
...
+                "swift"
+            ],
+            "name": "Release _TracingBenchmarks",
+            "program": "${workspaceFolder:swift-distributed-tracing}/.build/release/_TracingBenchmarks",
+            "args": [],
+            "cwd": "${workspaceFolder:swift-distributed-tracing}",
+            "preLaunchTask": "swift: Build Release _TracingBenchmarks"
+        }
+    ]
+}
```
stevapple, line 28: Better excluded since it's going to be generated? Maybe it's worth adding `.vscode/` into `.gitignore` list.

stevapple (changes requested):

.vscode/launch.json
adam-fowler: Ah that was a mistake

adam-fowler: removed

slashmo (approved):

ktoso (approved):
Sorry for the delay here was out and sick for a while, thanks! LGTM

And great to hear Hummingbird adopting!

ktoso:
@swift-server-bot add to allowlist

PR #132, opened 2023-08-11
Enable strict concurrency checking in CI

czechboy0:
## Summary

To help avoid concurrency bugs, this PR enables complete concurrency checking mode and addresses all the current concurrency warnings.

Some notable changes:
- Moved some locking primitives into SPI of Instrumentation, to be usable in the rest of the package
- Created a LockedValueBox type, used in several places to protect mutable state
- added sendability annotations where it made sense
- introduced a small workaround for accessing `stdout`/`stderr` (which emitted a warning, as it's a read-write global variable), using `fdopen`, which I think should be fine, but calling it out
- reformatted using `./scripts/validate_format.sh`, so might be worth reviewing the diff with `?w=1`

## Test Plan

Tested that all builds and tests pass on macOS and Linux.

### DIFF
--- a/Sources/Instrumentation/InstrumentationSystem.swift
+++ b/Sources/Instrumentation/InstrumentationSystem.swift
@@ -24,41 +24,71 @@ import ServiceContextModule
 /// ``instrument``: Returns whatever you passed to ``bootstrap(_:)`` as an ``Instrument``.
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) // for TaskLocal ServiceContext
 public enum InstrumentationSystem {
-    private static let lock = ReadWriteLock()
-    private static var _instrument: Instrument = NoOpInstrument()
-    private static var isInitialized = false
+    /// Marked as @unchecked Sendable due to the synchronization being
+    /// performed manually using locks.
+    private final class Storage: @unchecked Sendable {
+        private let lock = ReadWriteLock()
+        private var _instrument: Instrument = NoOpInstrument()
+        private var _isInitialized = false
+
+        func bootstrap(_ instrument: Instrument) {
+            self.lock.withWriterLock {
+                precondition(
+                    !self._isInitialized, """
+                    InstrumentationSystem can only be initialized once per process. Consider using MultiplexInstrument if
+                    you need to use multiple instruments.
+                    """
+                )
+                self._instrument = instrument
+                self._isInitialized = true
+            }
+        }
+
+        func bootstrapInternal(_ instrument: Instrument?) {
+            self.lock.withWriterLock {
+                self._instrument = instrument ?? NoOpInstrument()
+            }
+        }
+
+        var instrument: Instrument {
+            self.lock.withReaderLock { self._instrument }
+        }
+
+        func _findInstrument(where predicate: (Instrument) -> Bool) -> Instrument? {
+            self.lock.withReaderLock {
+                if let multiplex = self._instrument as? MultiplexInstrument {
+                    return multiplex.firstInstrument(where: predicate)
+                } else if predicate(self._instrument) {
+                    return self._instrument
+                } else {
+                    return nil
+                }
+            }
+        }
+    }
+
+    private static let shared = Storage()
 
     /// Globally select the desired ``Instrument`` implementation.
     ///
     /// - Parameter instrument: The ``Instrument`` you want to share globally within your system.
     /// - Warning: Do not call this method more than once. This will lead to a crash.
     public static func bootstrap(_ instrument: Instrument) {
-        self.lock.withWriterLock {
-            precondition(
-                !self.isInitialized, """
-                InstrumentationSystem can only be initialized once per process. Consider using MultiplexInstrument if
-                you need to use multiple instruments.
-                """
-            )
-            self._instrument = instrument
-            self.isInitialized = true
-        }
+        self.shared.bootstrap(instrument)
     }
 
     /// For testing scenarios one may want to set instruments multiple times, rather than the set-once semantics enforced by ``bootstrap(_:)``.
     ///
     /// - Parameter instrument: the instrument to boostrap the system with, if `nil` the ``NoOpInstrument`` is bootstrapped.
     internal static func bootstrapInternal(_ instrument: Instrument?) {
-        self.lock.withWriterLock {
-            self._instrument = instrument ?? NoOpInstrument()
-        }
+        self.shared.bootstrapInternal(instrument)
     }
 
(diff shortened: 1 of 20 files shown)


tomerd:
@swift-server-bot add to allowlist

Tests/TracingTests/ActorTracingTests.swift
```diff
@@ -27,13 +27,17 @@ final class ActorTracingTests: XCTestCase {
 func work() async {}
 
 actor Foo {
-    var bar = 0
+    let bar: LockedValueBox<Int> = .init(0)
```
ktoso, line 30: This part is why adding @Sendable to the operation closure is wrong, the withSpan must not be sendable and thus the `bar` does not have to do synchronization -- this would make tracing a complete pain in actors otherwise.

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -24,56 +24,78 @@ import ServiceContextModule
...
+                } else if predicate(self._instrument) {
+                    return self._instrument
+                } else {
+                    return nil
+                }
+            }
+        }
+    }
+
+    private static let shared = Storage()
```
ktoso, line 70: That's a nice change, thanks

Sources/Tracing/Tracer.swift
```diff
@@ -350,6 +350,53 @@ public func withSpan<T, Instant: TracerInstant>(
...
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) // for TaskLocal ServiceContext
+@preconcurrency
+public func withSpan<T>(
+    _ operationName: String,
+    context: @Sendable @autoclosure () -> ServiceContext = .current ?? .topLevel,
+    ofKind kind: SpanKind = .internal,
+    function: String = #function,
+    file fileID: String = #fileID,
+    line: UInt = #line,
+    _ operation: @Sendable (any Span) async throws -> T
```
ktoso, line 385: This must not be @Sendable -- see comment in test what this breaks then (e.g. inside an actor)

ktoso (changes requested):
Overall good, but we must not change the operation to be sendable, it breaks how tracing is intended to be used, comment here: https://github.com/apple/swift-distributed-tracing/pull/132/files#r1293083353

Once we revert changing `operation` closure types this should be okey. I'm also unsure why `context` closure would want to be sendable, seems iffy 🤔

czechboy0:
Thanks for the review, @ktoso, will address your feedback!

Tests/TracingTests/ActorTracingTests.swift
czechboy0: Actually, is the answer here to duplicate the `withSpan` methods in `extension Actor`, which implement the methods again using the building blocks of `startAnySpan` etc? That seems to make the warnings go away for me.

ktoso: But why would you want the operation to be sendable? We're always going to invoke it right away and never store it anywhere for "execute somewhere else"; they don't need to be sendable

czechboy0: I'm not saying they should be sendable :) I'm trying to fix the warnings I pasted above, and now exploring the idea of the extension on Actor.

ktoso: I'm a bit surprised the extension methods + Sendable does not cause warnings? Either way I think the answer here is just that operation closure just isn't sendable and that's ok 🆗  😄

czechboy0: The problem is that this test will then produce warnings when complete concurrency checking is enabled. And client code that tries to do this as well. I think we need _some_ solution here (even if making the closures isn't the right one), the code, as written, produces warnings with strict concurrency checking, today.

ktoso: We're backed into a corner here to be honest. Complete checking is strict, but the language does not offer right tools to solve this. The workaround (that even the stdlib uses) is done via `@_unsafeInheritExecutor` but that's incredibly unsafe and we cannot use it here --  we'd literarily be introducing real dataraces in correct looking code (this attribute removes ALL hop-to operations).

I'd be too worried to put it onto withSpan which is intended to wrap "entire functions" because it'll end up causing actual race conditions at runtime -- by now making any actor hops. 

I think we must wait it out and get to the safe version that would become `@inheritExecutor` and only then would complete checking be correct here... (and in the stdlib as well for that matter).

As potential workaround, I have explored the "overload" on Actor but it also hits a number of problems. We could do this: https://github.com/apple/swift-distributed-tracing/commit/9484f4ba1ad8ca43482666a7a24e3be23692eba9

I talked this over with @FranzBusch and @Lukasa a bunch, and end up coming back to the safe `@inheritExecutor` whenever we hit those issues. This should probably inform the priority of finding a solution for these issues as we plan for enabling complete checking in the near future.

FYI @DougGregor @hborla @tomerd

czechboy0: Understood - I'll back out the public API change and we should just land this PR with this warning, and address it separately. I wouldn't want to introduce new public API unless we want to maintain it long term.

Sources/Tracing/Tracer.swift
czechboy0: Reverted.

czechboy0:
@ktoso, ok I commented out the strict concurrency flags for now, but I think taking these changes would still be valuable. Once we have an answer for making `withSpan` safe, we can just uncomment the flags.

ktoso:
It's a good change over all, I'll give it another skim and merge.
Thank you @czechboy0 !

Related radars: rdar://113846321 rdar://113848512

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -24,56 +24,78 @@ import ServiceContextModule
...
     public static var instrument: Instrument {
-        self.lock.withReaderLock { self._instrument }
+        shared.instrument
     }
 }
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) // for TaskLocal ServiceContext
 extension InstrumentationSystem {
     /// INTERNAL API: Do Not Use
     public static func _findInstrument(where predicate: (Instrument) -> Bool) -> Instrument? {
```
ktoso, line 98: TODO: We can SPI this, I'll take that on :)

ktoso (approved):

ktoso:
Looks good, thank you @czechboy0 !

And we'll follow up on those radars about strict checking and closures

PR #133, opened 2023-09-06
Add Span.updateAttributes

adam-fowler:
Updating span attributes requires some form of thread synchronisation. If you update each attribute individually this could have a performance hit. This PR adds `Span.updateAttributes` which provides a way to update attributes in one go and only pay for the thread synchronisation costs once.


Sources/Tracing/SpanProtocol.swift
```diff
@@ -159,6 +159,23 @@ extension Span {
     }
 }
 
+extension Span {
+    /// Update Span attributes in a block instead of individually
```
slashmo, line 163:
```suggestion
    /// Update Span attributes in a block instead of individually.
```

Sources/Tracing/SpanProtocol.swift
```diff
@@ -159,6 +159,23 @@ extension Span {
...
+    ///
+    /// Updating a span attribute will involve some type of thread synchronisation
```
slashmo, line 165:
```suggestion
    /// Updating a span attribute will involve some type of thread synchronisation.
```

Sources/Tracing/SpanProtocol.swift
```diff
@@ -159,6 +159,23 @@ extension Span {
...
+    /// primitive to avoid multiple threads updating the attributes at the same
+    /// time. If you update each attributes individually this can cause slowdown.
+    /// This function updates the attributes in one call to avoid hitting the
+    /// thread synchronisation code multiple times
```
slashmo, line 169:
```suggestion
    /// thread synchronisation code multiple times.
```

slashmo (approved):
LGTM, thanks for opening the PR 👍 Tiny nitpicks in the doc blocks to keep it more consistent with the existing documentation.

PR #139, opened 2024-02-01
[PATCH] [bug] Fix ignored `at` argument in two instances of withSpan

ktoso:
In two cases, the signature for `withSpan` included the `at` parameter to override the starting timestamp on the span, but it was being ignored and `DefaultTracerClock.now` was being used instead.

resolves rdar://122048914


ktoso:
@swift-ci please smoke test

ktoso:

ktoso:
`.FileNotFoundError: [Errno 2] No such file or directory: './docker/docker-compose.2204.510.yaml'` need to make this I guess

PR #140, opened 2024-02-01
Remove old benchmark infra and replace with package-benchmark

ktoso:
Remove old benchmark infra which causes sendability warnings.

There's proper benchmark packages now and we should use them instead, depend on: https://github.com/ordo-one/package-benchmark.git

**Modifications:**
Remove custom benchmark infra.
Adopt existing benchmark package -- this can also nicely track allocations.

**Result:**

No sendability warnings; less deprecated benchmark infra


ktoso:
@swift-ci please smoke test

PR #144, opened 2024-04-01
update the otel-config to match latest otel image

heckj:
Found the existing configuration wasn't working with the latest otel docker image, and in digging, found that `jaeger` is no longer a valid exporter for otel-collector. Since Jaeger has an OTLP collector itself now directly available, this updates the 'jaeger' exporter to use `otlp` instead.

Kudos to
https://stackoverflow.com/questions/77475771/error-when-running-otel-collector-with-jaeger-in-docker-containers, which helped me identify the fix


slashmo (approved):
Thanks for the fix 🙏

PR #148, opened 2024-07-18
Adopt Swift 6.0 #isolation; Resolves async closure behavior in withSpan

ktoso:
**Motivation:**

This has been a problem for a long time, where we could not safely non sendable actor state from withSpan because the closure would "hop off". Now with #isolation we can handle it properly in Swift 6.0

**Modifications:**
Adopt #isolation parameter in all with... APIs with async closures.

**Result:**

Usages like this will not warn nor produce errors anymore:

```
actor Foo {
    var bar = 0

    func foo() async {
        var num = 0
         await withSpan(#function) { _ in
            // await self.withSpanWorkaround(#function) { _ in
            self.bar += 1
            await work()
            num += 1
        }
    }
}
```

resolves rdar://113848512

### DIFF
--- a/Sources/Tracing/Tracer.swift
+++ b/Sources/Tracing/Tracer.swift
@@ -322,7 +322,7 @@ public func withSpan<T>(
 ///   - operation: The operation that this span should be measuring
 /// - Returns: the value returned by `operation`
 /// - Throws: the error the `operation` has thrown (if any)
-#if swift(>=6.0.0)
+#if swift(>=6.0)
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) // for TaskLocal ServiceContext
 public func withSpan<T, Instant: TracerInstant>(
     _ operationName: String,
@@ -347,7 +347,10 @@ public func withSpan<T, Instant: TracerInstant>(
         try await operation(anySpan)
     }
 }
-#else
+#endif
+
+@_disfavoredOverload
+@available(*, deprecated, message: "Prefer #isolation version of this API")
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) // for TaskLocal ServiceContext
 public func withSpan<T, Instant: TracerInstant>(
     _ operationName: String,
@@ -371,7 +374,6 @@ public func withSpan<T, Instant: TracerInstant>(
         try await operation(anySpan)
     }
 }
-#endif
 
 /// Start a new ``Span`` and automatically end when the `operation` completes,
 /// including recording the `error` in case the operation throws.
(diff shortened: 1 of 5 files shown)


Sources/Tracing/Tracer.swift
```diff
@@ -315,18 +315,21 @@ public func withSpan<T>(
...
 /// - Returns: the value returned by `operation`
 /// - Throws: the error the `operation` has thrown (if any)
+#if swift(>=6.0.0)
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) // for TaskLocal ServiceContext
 public func withSpan<T, Instant: TracerInstant>(
     _ operationName: String,
     at instant: @autoclosure () -> Instant,
     context: @autoclosure () -> ServiceContext = .current ?? .topLevel,
     ofKind kind: SpanKind = .internal,
+    isolation: isolated (any Actor)? = #isolation,
```
0xTim, line 332: This parameter isn't documented now right?

0xTim (commented):
The doc comments could do with a look as they won't be consistent depending on your Swift version

ktoso:
I kept "old" API around but made it deprecated and disfavored in 6.0+.

I'll merge but if anyone has any concerns let me know and we'll follow up

Sources/Tracing/Tracer.swift
```diff
@@ -315,12 +315,42 @@ public func withSpan<T>(
...
 ///   - context: The `ServiceContext` providing information on where to start the new ``Span``.
 ///   - kind: The ``SpanKind`` of the new ``Span``.
+///   - isolation: Defaulted parameter for inheriting isolation of calling actor
 ///   - function: The function name in which the span was started
 ///   - fileID: The `fileID` where the span was started.
 ///   - line: The file line where the span was started.
 ///   - operation: The operation that this span should be measuring
 /// - Returns: the value returned by `operation`
 /// - Throws: the error the `operation` has thrown (if any)
+#if swift(>=6.0)
```
porglezomp, line 325: Is `#if swift(>=6.0)` correct here? Should this use `#if compiler(>=6.0.0)` instead, if a Swift 6 compiler in 5.x mode understands `#isolation`? I just tried to adopt this in pre-6 code with a 6.0 compiler and it was using the previous overload.

Sources/Tracing/Tracer.swift
```diff
@@ -315,12 +315,42 @@ public func withSpan<T>(
...
+        file: fileID,
+        line: line
+    ) { anySpan in
+        try await operation(anySpan)
+    }
+}
+#endif
+
+@_disfavoredOverload
+@available(*, deprecated, message: "Prefer #isolation version of this API")
```
porglezomp, line 353: Should the `deprecated` attribute be applied here when the `#isolation` API isn't available? Should this be in the same `#if` as above?

porglezomp (commented):
Post-merge review, I tried this out on the main branch and ran into lots of warnings.

ktoso:
Thanks @porglezomp lemme check this

PR #150, opened 2024-08-21
Fix use of Swift 6 language mode

orobio:
Adds a Package@swift-6.0.swift to make sure the package is compiled for the Swift 6 language mode. This fixes the visibility of the new withSpan functions, gated by `#if swift(>=6.0)`.

Not sure if this is the desired solution, but it currently works for us. Let me know if something else is required and I'll be happy to look into it.

Edit:
After digging in some more, I think it would be better to update all `#if swift(>=6.0)` conditionals to `#if compiler(>=6.0)`. Would that be an acceptable solution?

### DIFF
--- a/Sources/Tracing/Tracer.swift
+++ b/Sources/Tracing/Tracer.swift
@@ -322,7 +322,7 @@ public func withSpan<T>(
 ///   - operation: The operation that this span should be measuring
 /// - Returns: the value returned by `operation`
 /// - Throws: the error the `operation` has thrown (if any)
-#if swift(>=6.0)
+#if compiler(>=6.0)
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *) // for TaskLocal ServiceContext
 public func withSpan<T, Instant: TracerInstant>(
     _ operationName: String,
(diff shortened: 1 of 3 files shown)


Lukasa:
I think we should take your suggestion of updating to use `#if compiler`, not `#if swift`.

orobio:
> I think we should take your suggestion of updating to use `#if compiler`, not `#if swift`.

It's updated to use `#if compiler`. I just found a [comment](https://github.com/apple/swift-distributed-tracing/pull/148#discussion_r1704541776) on the previous pull request, regarding the deprecation being done unconditionally. I feel we should have both `@available...` and `@_disfavoredOverload` inside the `#if` blocks. If you agree, I can push an update for that as well.

Lukasa:
Yeah, I'm not wild about the deprecation being present when there is no alternative choice.

orobio:
> Yeah, I'm not wild about the deprecation being present when there is no alternative choice.

I pushed a fix for this as well.

Lukasa:
@ktoso I think we need a review from you here

slashmo (approved):
LGTM

ktoso (approved):
These look good, thank you very much!

Definitely agreed on #if guarding the deprecations as well, thank you for doing that

ktoso:
Thanks for the ping! Coming out of sick leave inbox was overwhelmed 😅

ktoso:

ktoso:
docc plugin seems borked on 5.7? Seems weird 🤔 Need to investigate what's up

jessezamora:
Anything new here?

orobio:
Could this have been a temporary glitch? Should we perhaps run the test again?

ktoso:

ktoso:
Sorry I was traveling and only now catching up here, let's see what's up here

ktoso:

ktoso:
Thank you, I'll cut a 1.1.2 with this

PR #153, opened 2024-11-08
Adds documentation on manual span propogation

NeedleInAJayStack:
**Motivation:**

I work with a number of packages that are heavily NIO-based. From the current documentation, I found it unclear how to create nested spans using this package in that environment where Task local variables were not preserved throughout EventLoopFuture chains.

**Modifications:**

This simply adds documentation for how to manually propogate spans by providing context arguments. It offers some examples that clarify that those contexts should come from started spans in order for trace IDs to be linked.

**Result:**

No functional changes, simply more thorough documentation.

Please let me know if I missed something or you have different suggestions!

### DIFF
--- a/Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
+++ b/Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
@@ -207,6 +207,37 @@ actor MySampleServer {
 
 While this code is very simple for illustration purposes, and it may seem surprising why there are two separate places where we need to call into user-code separately, in practice such situations can happen when using asynchronous network or database libraries which offer their API in terms of callbacks. Always consider if and when to restore context such that it makes sense for the end user.
 
+#### Manual propogation
+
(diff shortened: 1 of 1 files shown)


Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -207,6 +207,37 @@ actor MySampleServer {
 
 While this code is very simple for illustration purposes, and it may seem surprising why there are two separate places where we need to call into user-code separately, in practice such situations can happen when using asynchronous network or database libraries which offer their API in terms of callbacks. Always consider if and when to restore context such that it makes sense for the end user.
 
+#### Manual propogation
+
+There are circumstances where `task-local` variables are interrupted during normal execution flow. One common instance is when using
```
ktoso, line 212:
```suggestion
There are circumstances where [task-local variable](https://developer.apple.com/documentation/swift/tasklocal) propagation may be interrupted. One common instance is when using
```

ktoso (approved):
Sorry I missed this, I think it's a good addition

PR #157, opened 2024-11-22
Add a TracingMacros module providing the @Traced macro

### LINKED ISSUE #125: Offer @traced macro when macros gain function body replacement capabilities
Typically we trace at function level by wrapping the whole implementation in a span, as demonstrated by the sample.

```swift
func makeDinner() async throws -> Meal {
  try await withSpan("makeDinner") { _ in
    async let veggies = try chopVegetables()
    async let meat = marinateMeat()
    async let oven = preheatOven(temperature: 350)
    // ...
    return try await cook(veggies, meat, oven)
  }
}
```

It makes me think if we can have an `@Traced` macro attached to functions, which handles the span automatically for users. Eventually such macro should be configurable (explicit tracer, span name, extracting context from parameters, etc.), but AFAIK Swift macros seem not having such level of expressiveness yet.

```swift
@Traced
func makeDinner() async throws -> Meal {
  async let veggies = try chopVegetables()
  async let meat = marinateMeat()
  async let oven = preheatOven(temperature: 350)
  // ...
  return try await cook(veggies, meat, oven)
}
```

porglezomp:
This provides a `@Traced` macro which adds a tracing span around the body of the function. This macro allows customizing the `operationName`, the `context`, and the `kind` of the span, same as the `withSpan` function. It also exposes the span itself into the scope of the function, with a customizable name.

This doesn't attempt to support automatically adding parameters as attributes, because the scoping rules of attached macros aren't very amenable to controlling that. That could be added in the future if it's judged necessary.

## Examples:

```swift
@Traced("preheat oven")
func preheatOven(temperature: Int) async throws -> Oven {
    span.attributes["oven.targetTemperature"] = temperature
    await sleep(for: .seconds(6))
    return Oven()
}
```

expands to:

```swift
func preheatOven(temperature: Int) async throws -> Oven {
    withSpan("preheat oven") { span in
        span.attributes["oven.targetTemperature"] = temperature
        await sleep(for: .seconds(6))
        return Oven()
    }
}
```

## Notes:

This places the macros into a separate product so that users who don't want to pay the compile-time cost of macros don't have to use it, you opt-in to that cost by depending on the `TracingMacros` module.

This adds minimum OS versions for Apple platforms in order to be able to depend on SwiftSyntax. This applies to the whole package since there are no per-target platform specifications. Most notably: This raises the macOS minimum deployment target from the (implicit) 10.13 to 10.15 (matching SwiftSyntax).

Resolves #125

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,6 +1,6 @@
 // swift-tools-version:5.9
-import PackageDescription
 import CompilerPluginSupport
+import PackageDescription
 
 let package = Package(
     name: "swift-distributed-tracing",
--- a/Package.swift
+++ b/Package.swift
@@ -1,14 +1,24 @@
 // swift-tools-version:5.9
 import PackageDescription
+import CompilerPluginSupport
 
 let package = Package(
     name: "swift-distributed-tracing",
+    platforms: [
+        .macOS(.v10_15),
+        .iOS(.v13),
+        .tvOS(.v13),
+        .watchOS(.v6),
+        .macCatalyst(.v13),
+    ],
     products: [
         .library(name: "Instrumentation", targets: ["Instrumentation"]),
         .library(name: "Tracing", targets: ["Tracing"]),
+        .library(name: "TracingMacros", targets: ["TracingMacros"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-service-context.git", from: "1.1.0")
+        .package(url: "https://github.com/apple/swift-service-context.git", from: "1.1.0"),
+        .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "600.0.0-latest"),
     ],
     targets: [
         // ==== --------------------------------------------------------------------------------------------------------
@@ -41,6 +51,33 @@ let package = Package(
             name: "TracingTests",
             dependencies: [
                 .target(name: "Tracing")
+            ]
+        ),
+
+        // ==== --------------------------------------------------------------------------------------------------------
+        // MARK: TracingMacros
+
+        .target(
+            name: "TracingMacros",
+            dependencies: [
+                .target(name: "Tracing"),
+                .target(name: "TracingMacrosImplementation"),
+            ]
+        ),
+        .macro(
+            name: "TracingMacrosImplementation",
+            dependencies: [
+                .product(name: "SwiftSyntaxMacros", package: "swift-syntax"),
+                .product(name: "SwiftCompilerPlugin", package: "swift-syntax"),
+            ]
+        ),
+        .testTarget(
+            name: "TracingMacrosTests",
+            dependencies: [
+                .target(name: "Tracing"),
+                .target(name: "TracingMacros"),
+                .target(name: "TracingMacrosImplementation"),
+                .product(name: "SwiftSyntaxMacrosTestSupport", package: "swift-syntax"),
             ]
         ),
     ]
--- a/Sources/TracingMacros/TracedMacro.swift
+++ b/Sources/TracingMacros/TracedMacro.swift
@@ -15,11 +15,26 @@
 import Tracing
 
 #if compiler(>=6.0)
+/// Instrument a function to place the entire body inside a span.
+///
+/// This macro is equivalent to calling ``/Tracing/withSpan`` in the body, but saves an
+/// indentation level and duplication. It introduces a `span` variable into the
+/// body of the function which can be used to add attributes to the span.
+///
+/// Parameters are passed directly to ``/Tracing/withSpan`` where applicable,
+/// and omitting the parameters from the macro omit them from the call, falling
+/// back to the default.
+///
+/// - Parameters:
+///   - operationName: The name of the operation being traced. Defaults to the name of the function.
+///   - context: The `ServiceContext` providing information on where to start the new ``/Tracing/Span``.
+///   - kind: The ``/Tracing/SpanKind`` of the new ``/Tracing/Span``.
+///   - spanName: The name of the span variable to introduce in the function. Pass `"_"` to omit it.
 @attached(body)
 public macro Traced(
     _ operationName: String? = nil,
     context: ServiceContext? = nil,
     ofKind kind: SpanKind? = nil,
-    span spanName: String? = nil
+    span spanName: String = "span"
 ) = #externalMacro(module: "TracingMacrosImplementation", type: "TracedMacro")
 #endif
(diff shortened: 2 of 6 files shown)


porglezomp:
(Sorry @Traced that the commit messages tag you, I didn't mean to do that! 🙇🏻‍♀️)

porglezomp:
My bad with the trailing commas, I tested on 5.9 but not on 6.0, and I was using a toolchain with https://github.com/swiftlang/swift-evolution/pull/2344

Sources/TracingMacros/TracedMacro.swift
```diff
@@ -0,0 +1,25 @@
...
+// See CONTRIBUTORS.txt for the list of Swift Distributed Tracing project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+@_exported import ServiceContextModule
+import Tracing
+
+#if compiler(>=6.0)
+@attached(body)
```
ktoso, line 18: Needs some docs.

The shape is good though! I like the `spanName` as well, that's quite niffty.

Sources/TracingMacros/TracedMacro.swift
```diff
@@ -0,0 +1,25 @@
...
+public macro Traced(
+    _ operationName: String? = nil,
```
ktoso, line 20: I wonder if this may want to be an `TracedOperationName` like this:

```
public struct TracedOperationName: ExpressibleByStringLiteral{ 
let value: TracedOperationName

package enum TracedOperationName {
case baseName // the default "foo"
case fullName // "foo(bar:)
case string(String)
}


public static var baseName: ... // for all the above
```

WDYT?

Sources/TracingMacros/TracedMacro.swift
```diff
@@ -0,0 +1,25 @@
...
+    context: ServiceContext? = nil,
```
ktoso, line 21: Hmm, technically is the default `.current` rather than `nil`?

Should the type be the `@autoclosure () -> ServiceContext` as well, so we avoid reading the local when tracing is disabled (not bootstrapped)

ktoso (commented):
Very nice! I had some minor comments, please have a look

Sources/TracingMacros/TracedMacro.swift
porglezomp: I was going back and forth on whether it should be implemented with `baseName` / `fullName` semantics, so this looks good to me! Should this type be in `TracingMacros`?

Sources/TracingMacros/TracedMacro.swift
porglezomp: I'm not entirely sure about the answer here since this is my first major macro.
- For the default: I'm using `nil` just to delegate the default handling to `withSpan` itself—we don't even construct the syntax to pass a parameter when it's omitted here. I don't fully understand how default parameters are even passed to the macro.
- For the type: since this is a macro it's handled via syntax expansion, so the type is used for checking the interface, not as a runtime type. When the expression is interpolated into the call, I think it will end up becoming an autoclosure at the `withSpan` callsite?

porglezomp: I did some experimentation: The default value isn't passed to the macro at all, it seems to just be for documentation purposes?

porglezomp: Since it seems to be only documentation—do we want to duplicate the default values from `withSpan` here, or cross-reference to say "look at the default values in the function documentation"? I can see arguments for both. I leaned towards omitting them here to not duplicate code across modules, but it's already duplicated across many overload so it's probably worth it for the documentation's sake.

Sources/TracingMacros/TracedMacro.swift
ktoso: Yup, right next to this `Traced` declaration would be fine

Sources/TracingMacros/TracedMacro.swift
ktoso: Oh good point about the autoclosure not mattering! 

the nil maybe is fine after all, since it just means "don't pass anything" makes sense to me

For the docs: i think it's fine to not duplicate the docs, just a short note that it is passed to the `withSpan` will be enough

Sources/TracingMacros/TracedMacro.swift
porglezomp: Actually I just realized this is also made slightly harder by the macro—we want this conceptually but we only have access to the syntax and not the evaluated value. Two options here:
- Directly match looking for `.baseName` / `.fullName` / string expression, or add a bit of runtime support expanding to something like:

```swift
withSpan(.baseName.text(baseName: "foo", fullName: "foo(bar:)")) { span in ... }
```

And then doing the selection in regular code.

Package.swift
```diff
@@ -1,14 +1,24 @@
...
 
 let package = Package(
     name: "swift-distributed-tracing",
+    platforms: [
+        .macOS(.v10_15),
+        .iOS(.v13),
+        .tvOS(.v13),
+        .watchOS(.v6),
+        .macCatalyst(.v13),
+    ],
```
Lukasa, line 13: This change is a semver major, are we actually intending to mint a new semver for this?

Lukasa (changes requested):

Sources/TracingMacros/TracedMacro.swift
ktoso: I see, yeah let's document to just look at the real method maybe

Sources/TracingMacros/TracedMacro.swift
ktoso: Yes, it would be in TracingMacros I think. I think basename as basic is okey

Package.swift
Lukasa: Yeah, unfortunately this is a breaking change because SwiftPM doesn’t take the minimum deployment targets into consideration when resolving versions. The result is broken builds, caused by picking up the new minima. 

While we’re on the topic of breakage, Swift-syntax also represents a risk. Because of their versioning strategy it’s very easy to produce unsolveable build graphs, or forcing users to not be able to update past the current version. I’m generally nervous about putting this anywhere it might accidentally get picked up.

porglezomp: For that at least—am I correct that package versions are only resolved for products that are dependent on, so that at least requires libraries to depend on the macros to get breakage? But yes I don't have any experience with releasing a package providing macros, I have no idea what is sufficient here. Just putting it in `-extras`? Putting it in a whole separate package?

ktoso: I think extras might be fine -- this was replaced by https://github.com/apple/swift-distributed-tracing-extras/pull/42

ktoso:
Replaced by: in https://github.com/apple/swift-distributed-tracing-extras/pull/42 -extras

PR #160, opened 2024-12-16
Add recordingSpan and currentSpan methods to TracerProtocol

slashmo:
**Motivation:**

Currently, it is only possible to implicitly work with the current span by transparently creating a child span (using ServiceContext.current) under the hood. This is sufficient for almost all use-cases, but does not work in case a piece of code wants to add an event to the current span without having a handle on said span, e.g. if the span was created by a library.

The concrete use-case I'm facing is building a tracing "hook" for [my `OpenFeature` client](https://github.com/swift-open-feature/swift-open-feature). `OpenFeature` hooks are similar to middleware, but instead of wrapping a piece of logic they are called before and after a feature flag was evaluated. Therefore, `OpenFeature` implementations in other ecosystems have opted to only add an event instead of creating a new span for the flag evaluation itself:

- [OpenFeature Go SDK: Tracing Hook](https://github.com/open-feature/go-sdk-contrib/blob/629a08258078fc7ee6414390f8bce4faae44c8d2/hooks/open-telemetry/pkg/traces.go#L55)
- [OpenFeature JS SDK: Tracing Hook](https://github.com/open-feature/js-sdk-contrib/blob/e773416a4a0dd67db8e5a79a5ffb2d95b3295714/libs/hooks/open-telemetry/src/lib/traces/tracing-hook.ts#L21)

**Modifications:**

I added a `recordingSpan(identifiedBy context: ServiceContext) -> Span?` requirement to `TracerProtocol`. This way, `Tracer` implementations may look up and return a span identified by the data they stored in the provided `ServiceContext`. It's worth noting that this method is only intended for obtaining spans which are still recording as opposed to already ended ones that may not even be in memory anymore. I also added a default implementation of this method to avoid introducing a breaking change. On top of this new protocol requirement, I added an extension to `TracerProtocol` with sugar to obtain the current span based on the task-local `ServiceContext`.

**Result:**

Library authors and application developers are now able to look up the current recording span to interact with it by e.g. adding events and attributes:

```swift
func spanCreatingMethod() async {
    await withSpan("operation") {
        await methodWithoutSpanArgument()
    }
}

func methodWithoutSpanArgument() async {
    if let span = InstrumentationSystem.tracer.currentSpan() {
        span.addEvent("example")
    }
}
```

### DIFF
--- a/Sources/Tracing/TracerProtocol.swift
+++ b/Sources/Tracing/TracerProtocol.swift
@@ -66,7 +66,7 @@ public protocol Tracer: LegacyTracer {
     /// It was added retroactively with a default implementation returning `nil` and therefore isn't guaranteed to be implemented by all `Tracer`s.
     /// - Parameter context: The context containing information that uniquely identifies the span being obtained.
     /// - Returns: The span identified by the given `ServiceContext` in case it's still recording.
-    func recordingSpan(identifiedBy context: ServiceContext) -> Span?
+    func activeSpan(identifiedBy context: ServiceContext) -> Span?
 }
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)  // for TaskLocal ServiceContext
@@ -115,12 +115,12 @@ extension Tracer {
         )
     }
 
-    /// Default implementation for ``recordingSpan(identifiedBy:)`` which always returns `nil`.
-    /// This default exists in order to facilitate source-compatible introduction of the ``recordingSpan(identifiedBy:)`` protocol requirement.
+    /// Default implementation for ``activeSpan(identifiedBy:)`` which always returns `nil`.
+    /// This default exists in order to facilitate source-compatible introduction of the ``activeSpan(identifiedBy:)`` protocol requirement.
     ///
     /// - Parameter context: The context containing information that uniquely identifies the span being obtained.
     /// - Returns: `nil`.
-    public func recordingSpan(identifiedBy context: ServiceContext) -> Span? {
+    public func activeSpan(identifiedBy context: ServiceContext) -> Span? {
         nil
     }
 }
--- a/Sources/Tracing/TracerProtocol.swift
+++ b/Sources/Tracing/TracerProtocol.swift
@@ -123,15 +123,6 @@ extension Tracer {
     public func recordingSpan(identifiedBy context: ServiceContext) -> Span? {
         nil
     }
-
-    /// Attempt to retrieve the current recording span based on the task-local `ServiceContext`.
-    ///
-    /// - Note: This method should be considered best-effort as it might not (yet) be supported by the application author's `Tracer` of choice.
-    /// - Returns: A span if one can be obtained via the task-local `ServiceContext`.
-    public func currentSpan() -> Span? {
-        guard let context = ServiceContext.current else { return nil }
-        return recordingSpan(identifiedBy: context)
-    }
 }
 
 // ==== ----------------------------------------------------------------------------------------------------------------
--- a/Sources/Tracing/TracerProtocol.swift
+++ b/Sources/Tracing/TracerProtocol.swift
@@ -59,6 +59,14 @@ public protocol Tracer: LegacyTracer {
         file fileID: String,
         line: UInt
     ) -> Self.Span
+
+    /// Retrieve the recording span for the given `ServiceContext`.
+    ///
+    /// - Note: This API does not enable look up of already finished spans.
+    /// It was added retroactively with a default implementation returning `nil` and therefore isn't guaranteed to be implemented by all `Tracer`s.
+    /// - Parameter context: The context containing information that uniquely identifies the span being obtained.
+    /// - Returns: The span identified by the given `ServiceContext` in case it's still recording.
+    func recordingSpan(identifiedBy context: ServiceContext) -> Span?
 }
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)  // for TaskLocal ServiceContext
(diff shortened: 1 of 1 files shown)


slashmo:
Here's a draft implementation of the `ServiceContext`-based recording span lookup in Swift OTel: https://github.com/swift-otel/swift-otel/pull/167

czechboy0:
This looks very useful, thanks @slashmo!

Do I understand correctly that this method needs to be considered "best effort", since existing implementations might not be implementing it (yet, or ever) - so whether new events attached this way will show up depends on the concrete backend? It's fine IMO, we might just want to call this out explicitly. (It'll be the case with any new-and-defaulted requirements like this one, but it might not be obvious to new adopters of swift-distributed-tracing.)

slashmo:
> Do I understand correctly that this method needs to be considered "best effort".

Yes, it would have to be. Thanks for bringing it up. This reminds me a bit of when we introduced metadata providers in `Logging`, which also provides a default no-op implementation for backwards-compatibility but also includes a one-off debug log to indicate that the default is being used if the bootstrapped handler does not support metadata providers (yet):

https://github.com/apple/swift-log/blob/6efb781a2900883e1c702ad8b782b14dd1c75c13/Sources/Logging/LogHandler.swift#L181

```swift
extension LogHandler {
    /// Default implementation for `metadataProvider` which defaults to `nil`.
    /// This default exists in order to facilitate source-compatible introduction of the `metadataProvider` protocol requirement.
    public var metadataProvider: Logger.MetadataProvider? {
        get {
            nil
        }
        set {
            #if DEBUG
            if LoggingSystem.warnOnceLogHandlerNotSupportedMetadataProvider(Self.self) {
                self.log(
                    level: .warning,
                    message:
                        "Attempted to set metadataProvider on \(Self.self) that did not implement support for them. Please contact the log handler maintainer to implement metadata provider support.",
                    metadata: nil,
                    source: "Logging",
                    file: #file,
                    function: #function,
                    line: #line
                )
            }
            #endif
        }
    }
}
```

Perhaps we could borrow this approach here as well? Either way, I agree that we should highlight it a bit more in documentation as well.

slashmo:
Hm, just realized that here we'd not be able to use `Logging` for the one-off warning because `swift-distributed-tracing` doesn't depend on `swift-log`. So either we'd have to `print` the debug log or find an alternative approach.

czechboy0:
I'm fine with us calling it out in the doc comments only. We can reopen the discussion of whether tracing should depend on logging separately, I don't think it needs to block this.

slashmo:
Sounds good. I'll update the docs then to provide some more clarity.

ktoso:
Give me a moment to come back to this after holidays, I'm back after the 13th. Technically we didn't do these methods on purpose but I can see the points being mentioned. I'd like to think this through properly. Thanks

Sources/Tracing/TracerProtocol.swift
```diff
@@ -59,6 +59,14 @@ public protocol Tracer: LegacyTracer {
...
         line: UInt
     ) -> Self.Span
+
+    /// Retrieve the recording span for the given `ServiceContext`.
+    ///
+    /// - Note: This API does not enable look up of already finished spans.
+    /// It was added retroactively with a default implementation returning `nil` and therefore isn't guaranteed to be implemented by all `Tracer`s.
+    /// - Parameter context: The context containing information that uniquely identifies the span being obtained.
+    /// - Returns: The span identified by the given `ServiceContext` in case it's still recording.
+    func recordingSpan(identifiedBy context: ServiceContext) -> Span?
```
ktoso, line 69: I think this is okey, but I'd perhaps rename it. Names used by the linked libraries sound ok to me, especially `activeSpan(identifiedBy: context)` seems pretty good. 

I'm not sure about the 'record**ing**` because it makes it sound like not a getter somehow

Sources/Tracing/TracerProtocol.swift
```diff
@@ -106,6 +114,24 @@ extension Tracer {
...
+    }
+
+    /// Attempt to retrieve the current recording span based on the task-local `ServiceContext`.
+    ///
+    /// - Note: This method should be considered best-effort as it might not (yet) be supported by the application author's `Tracer` of choice.
+    /// - Returns: A span if one can be obtained via the task-local `ServiceContext`.
+    public func currentSpan() -> Span? {
+        guard let context = ServiceContext.current else { return nil }
+        return recordingSpan(identifiedBy: context)
+    }
```
ktoso, line 134: I'm very on the fence about this sugar method. Let's keep sugar out of the API package as much as we can tbh... I'm worried about folks using this a lot and then suddenly not entering info into a span because it's not active anymore 🤔 

If people hit this a lot we could reconsider though but so far I've not seen many requests for it hm. Can we make this a separate issue and wait it out a bit?

ktoso (commented):
Minor comments, I think the getting span is ok but I'm worried about the sugar encouraging brittle patterns -- maybe let's accept without the sugar for now, ok?

Sources/Tracing/TracerProtocol.swift
ktoso: Or if we really really want it, I wonder if this is `Tracer.activeSpan` or `Tracer.currentActiveSpan` -- note that as properties, not functions.

Sources/Tracing/TracerProtocol.swift
slashmo: > I'm not sure about the 'recording` because it makes it sound like not a getter somehow

Yeah, good point. I haven't read it this way before, but I can see how it might be misunderstood. What about using `activeSpan(identifiedBy:)` here instead?

Sources/Tracing/TracerProtocol.swift
slashmo: Yeah, I can see how this could be accidentally "overused". I'm happy to remove it to make it purposefully harder to use the API.

slashmo: Removed ✅

Sources/Tracing/TracerProtocol.swift
slashmo: I renamed it to `activeSpan(identifiedBy:)` now, LGTM.

ktoso (approved):

ktoso:
Thanks! Let's take it a step at a time with that one then, LGTM!

Sources/Tracing/TracerProtocol.swift
```diff
@@ -59,6 +59,14 @@ public protocol Tracer: LegacyTracer {
         file fileID: String,
         line: UInt
     ) -> Self.Span
+
+    /// Retrieve the recording span for the given `ServiceContext`.
```
czechboy0, line 63: s/recording/active

slashmo: 😬 Thanks for noticing. I'll create a follow-up to rename the comment.

slashmo: Fixed in #164

PR #173, opened 2025-06-18
Tracer.withSpan's isolation param should be 'isolated'

ktoso:
**Motivation:**

The intent of this parameter is to not cross an isolation boundary when calling withSpan, thus, we were missing the isolated modifier in this protocol.

**Modifications:**

Make the `isolation: (any Actor)? = #isolation,` parameter 'isolated', as was originally intended

**Result:**

Not crossing actor boundary when calling this method means less errors when adopting Swift 6 mode, and correct behavior of these functions


czechboy0 (approved):

ktoso:
Gonna tag a patch with this :)

PR #174, opened 2025-06-23
feat: Add swift wasm compilation support to swift-distributed-tracing

scottmarchant:
# Summary

This PR adds support for compiling swift-distributed-tracing to wasm using the [SwiftWasm](https://swiftwasm.org) sdk.

This PR is [part of a larger effort](https://github.com/PassiveLogic/swift-web-examples/issues/1) by a company called PassiveLogic to enable broad support for SwiftWasm.

# Details

There are three changes required to enable wasm compilation for swift-distributed-tracing.

- Removed unused `import Dispatch` imports. SwiftWasm doesn't currently include GCD. Luckily, all usage of GCD was unused.
- Added some missing wasi-libc and pthread imports
- Add shim to allow wrapped access to CLOCK_REALTIME struct from WASILibc. This is a critical piece of distributed tracing to acquire timestamps that needed some manual shim code to map to the matching API in WASI.

# Testing done

- [x] - Cleaned up swiftformat lint on modified lines of change.
- [x] - Verified unit tests still pass with these changes
- [x] - Verified `swift build` completes without errors
- [x] - Verified `swift build --swift-sdk wasm32-unknown-wasi` completes without errors
- [x] - Verified a third-party executable can build this library as part of a larger wasm executable using the command `swift package --swift-sdk wasm32-unknown-wasip1-threads js --use-cdn`

# Impact Risk

Realistically, this change should not cause any impact or risk. Outside of wasm targets, the changes in this PR are additive.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -53,7 +53,6 @@ let package = Package(
             name: "_CWASI",
             dependencies: []
         ),
-
     ]
 )
 
--- a/Package.swift
+++ b/Package.swift
@@ -35,6 +35,7 @@ let package = Package(
             dependencies: [
                 .product(name: "ServiceContextModule", package: "swift-service-context"),
                 .target(name: "Instrumentation"),
+                .target(name: "_CWASI", condition: .when(platforms: [.wasi])),
             ]
         ),
         .testTarget(
@@ -43,6 +44,16 @@ let package = Package(
                 .target(name: "Tracing")
             ]
         ),
+
+        // ==== --------------------------------------------------------------------------------------------------------
+        // MARK: Wasm Support
+
+        // Provides C shims for compiling to wasm
+        .target(
+            name: "_CWASI",
+            dependencies: []
+        ),
+
     ]
 )
 
--- a/Sources/Tracing/TracingTime.swift
+++ b/Sources/Tracing/TracingTime.swift
@@ -23,8 +23,14 @@ import Glibc
 import Android
 #elseif canImport(Musl)
 import Musl
+#elseif canImport(WASILibc)
+import WASILibc
 #else
 #error("Unsupported runtime")
+#endif
+
+#if canImport(_CWASI)
+import _CWASI
 #endif
 
 public protocol TracerInstant: Comparable, Hashable, Sendable {
--- a/Sources/Tracing/TracingTime.swift
+++ b/Sources/Tracing/TracingTime.swift
@@ -90,7 +90,11 @@ public struct DefaultTracerClock {
 
     public var now: Self.Instant {
         var ts = timespec()
-        clock_gettime(CLOCK_REALTIME, &ts)
+        #if os(WASI)
+            CWASI_clock_gettime_realtime(&ts)
+        #else
+            clock_gettime(CLOCK_REALTIME, &ts)
+        #endif
         /// We use unsafe arithmetic here because `UInt64.max` nanoseconds is more than 580 years,
         /// and the odds that this code will still be running 530 years from now is very, very low,
         /// so as a practical matter this will never overflow.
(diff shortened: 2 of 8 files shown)


Sources/Tracing/TracingTime.swift
```diff
@@ -23,10 +23,16 @@ import Glibc
 import Android
```
scottmarchant, line 23: Note, there is a lot of existing swiftformat lint in this repository, including the existing indentation on many of these import blocks. I fixed the lint on the lines of code I'm changing in this PR. But I didn't fix it for everything else, to avoid making the diff far too large.

MaxDesiatov:
@FranzBusch @ktoso would either of you have a moment to review?

Sources/_CWASI/_CWASI.c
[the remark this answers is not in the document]
MaxDesiatov: We need this as a workaround for an empty target that only has include files correctly that SwiftPM can't handle otherwise.

scottmarchant:
> @scottmarchant could you update PR description to refer to official swift.org Swift SDKs for Wasm instead of SwiftWasm? Thanks!

@MaxDesiatov Done

Sources/_CWASI/_CWASI.c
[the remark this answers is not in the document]
scottmarchant: Yep, just needed to avoid issues with empty SPM targets, since the rest of the code in `_CWASI` is in the headers.

Package.swift
```diff
@@ -35,6 +35,7 @@ let package = Package(
             dependencies: [
                 .product(name: "ServiceContextModule", package: "swift-service-context"),
                 .target(name: "Instrumentation"),
+                .target(name: "_CWASI", condition: .when(platforms: [.wasi])),
```
FranzBusch, line 38: We seem to be getting an error on this line on 5.10

> error: Target Tracing imports another target (_CWASI) in the package without declaring it a dependency.

Is this a bug in the package manager on 5.10? Can you take a look and if it is a bug we can just disable the strict dependency check on just 5.10

Sources/Tracing/TracingTime.swift
```diff
@@ -23,10 +23,16 @@ import Glibc
...
 #elseif canImport(Musl)
 import Musl
+#elseif canImport(WASILibc)
+    import WASILibc
 #else
 #error("Unsupported runtime")
 #endif
 
+#if canImport(_CWASI)
```
MaxDesiatov, line 32:
```suggestion
#if os(WASI)
```

MaxDesiatov:
@scottmarchant would you mind addressing formatting job failures per the build log? https://github.com/apple/swift-distributed-tracing/actions/runs/16578688393/job/46920017078?pr=174

Sources/Tracing/TracingTime.swift
scottmarchant: @MaxDesiatov How about I do `#if os(WASI) && canImport(_CWASI)`? In theory all three options would have the same effect. Just want to preserve the canImport since it is conditionally available.

Package.swift
scottmarchant: @FranzBusch Indeed it does look like that is a bug. I tracked it down to `FB14859516` metnioned in https://forums.swift.org/t/do-swiftpm-conditional-target-dependencies-work/74047.

I removed the dependency check for swift versions before Swift 6. See https://github.com/apple/swift-distributed-tracing/pull/174/commits/386e324dbb9b6f3d3fe3eb0244ca6d96ba6fec63

scottmarchant:
> Can you fix the YAML lint and update against the latest main branch. Thanks again for working on this!

@FranzBusch Done. And thank you for the active reviews here! Hugely appreciated!

scottmarchant:
@FranzBusch I believe this branch has the latest main merged into it and is ready for merge now once the workflow runs.

Issue #165, opened 2025-02-08
Cannot checkout repository on windows because of file names with ":" in it


bhirt:
I have a windows swift project that recently had swift-distributed-tracing included as a dependency from some other dependency.   However, since there are files in this repository with a colon ':' in the filename, the repo cannot be checked out.  NTFS does not allow ":" in file/path names.

I haven't looked too much into this, and I'm not sure if it's something this team is interested in addressing.  It may be good practice to not use ":" in filenames, especially since there is a growing windows swift community -- and not using a colon is pretty minor inconvenience.  

```
PS C:\tmp> git clone https://github.com/apple/swift-distributed-tracing.git
Cloning into 'swift-distributed-tracing'...
remote: Enumerating objects: 3437, done.
remote: Counting objects: 100% (780/780), done.
remote: Compressing objects: 100% (286/286), done.
remote: Total 3437 (delta 675), reused 497 (delta 494), pack-reused 2657 (from 1)
Receiving objects: 100% (3437/3437), 2.01 MiB | 4.26 MiB/s, done.
Resolving deltas: 100% (2206/2206), done.
error: invalid path 'Benchmarks/Thresholds/5.10/TracingBenchmarks.NoopTracing.attribute:_set,_span.attributes.http.status_code_=_200.p90.json'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
PS C:\tmp>
```

Thanks for your time.

bhirt:
It looks like a couple changes in Benchmarks/Benchmarks/TracingBenchmarks/AtrributeDSLBenchmark.swift, along with renaming existing files in the repo would address this for future releases.     I could probably make a pull request if this is something the teams wants to address, and I knew what character ":" would be replaced with.

slashmo:
Thanks for flagging this problem and for investigating! Yep, I came to the same conclusion now. We can totally refrain from using colons inside benchmark names. A PR fixing the names would be great. I think `NoopTracing.attribute.set` instead of `NoopTracing.attribute: set` would be a good option here.

I'll also file an issue for [ordo-one/package-benchmark](https://github.com/ordo-one/package-benchmark) to start a discussion on whether creating problematic file paths could be prevented in the first place.

czechboy0:
Hi @bhirt,

[...]

PR #176, opened 2025-07-12
removing colons for windows cloning support

zamderax:
Addresses #165 so that Windows users can clone this library.

1. Replaced `:` with `.` 
2. Renamed the json files to use periods `.` instead of `:`


ktoso (approved):
Thanks for the patch, that’s fine.

PR #178, opened 2025-08-04
Revive TracingMacros and @Traced

### LINKED ISSUE #125: Offer @traced macro when macros gain function body replacement capabilities

ktoso:
This brings back, and will polish up a bit the `@Traced` macro that was initially pitched over here #157 by @porglezomp.

I'm signing up for long term maintenance of the macro, and we need to post about expectations and rules of macro use/adoption in libraries, but we should not outright ban them.

----

> This provides a `@Traced` macro which adds a tracing span around the body of the function. This macro allows customizing the `operationName`, the `context`, and the `kind` of the span, same as the `withSpan` function. It also exposes the span itself into the scope of the function, with a customizable name.
> 
> This doesn't attempt to support automatically adding parameters as attributes, because the scoping rules of attached macros aren't very amenable to controlling that. That could be added in the future if it's judged necessary.
> 
> ## Examples:
> 
> ```swift
> @Traced("preheat oven")
> func preheatOven(temperature: Int) async throws -> Oven {
>     span.attributes["oven.targetTemperature"] = temperature
>     await sleep(for: .seconds(6))
>     return Oven()
> }
> ```
> 
> expands to:
> 
> ```swift
> func preheatOven(temperature: Int) async throws -> Oven {
>     withSpan("preheat oven") { span in
>         span.attributes["oven.targetTemperature"] = temperature
>         await sleep(for: .seconds(6))
>         return Oven()
>     }
> }
> ```
> 

Resolves #125

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,6 +1,6 @@
 // swift-tools-version:5.9
-import PackageDescription
 import CompilerPluginSupport
+import PackageDescription
 
 let package = Package(
     name: "swift-distributed-tracing",
--- a/Package.swift
+++ b/Package.swift
@@ -1,14 +1,24 @@
 // swift-tools-version:5.9
 import PackageDescription
+import CompilerPluginSupport
 
 let package = Package(
     name: "swift-distributed-tracing",
+    platforms: [
+        .macOS(.v10_15),
+        .iOS(.v13),
+        .tvOS(.v13),
+        .watchOS(.v6),
+        .macCatalyst(.v13),
+    ],
     products: [
         .library(name: "Instrumentation", targets: ["Instrumentation"]),
         .library(name: "Tracing", targets: ["Tracing"]),
+        .library(name: "TracingMacros", targets: ["TracingMacros"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-service-context.git", from: "1.1.0")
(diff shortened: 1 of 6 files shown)


ktoso:
Heh, I forgot about the trouble with platform bump here... We may put it into a separate package or need to re-consider bumping the platforms here and making it a major version bump hm...

spprichard:
What's the current status on this? Is it still something worth pursuing?

PR #193, opened 2025-10-23
Drop Swift 5.9 and 5.10 support

kukushechkin:
Make Swift 6.0 the minimum support version.

### Motivation:

`swift-distributed-tracing` supports 3 latest Swift releases, which is at this point 6.0, 6.1 and 6.2. Swift 5 is no longer supported.

### Modifications:

- Package.swift swift tools set to 6.0
- CI configs no longer include Swift 5 and include 6.0, 6.1 and 6.2 for all supported platforms
- Removed checks if compiler is <6.0

### Result:

Swift 5 is no longer supported.


czechboy0 (approved):
Thanks 🙏

PR #196, opened 2025-10-29
Avoid testing global InstrumentationSystem state

kukushechkin:
Refactor tests to only use `InstrumentationSystem` global state bootstrapping where necessary.

### Motivation:

Decoupling tests from the global state allows running them in parallel later on within Swift Testing.

### Modifications:

- Use local `tracer` variable instead of calling `InstrumentationSystem.bootstrapInternal()`
- Add tests specifically for global `withSpan()`-like methods

### Result:

Most of the tests no longer rely on the global `InstrumentationSystem` state. Test coverage is up slightly.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,6 +43,10 @@ jobs:
     with:
       build_scheme: swift-distributed-tracing-Package
 
+  wasm-sdk:
+    name: WebAssembly Swift SDK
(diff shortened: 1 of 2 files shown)


ktoso (approved):
Cool, lgtm 👍

ktoso:
Technically this is "semver/none" not patch I guess, no real functionality touched at all

PR #198, opened 2025-11-06
fixing docs 

### LINKED ISSUE #182: Ensure all public APIs have docs
Enable AllPublicDeclarationsHaveDocumentation in the swift-format config file.

Add `// swift-format-ignore: AllPublicDeclarationsHaveDocumentation` when a comment is inherited from a declaration elsewhere.

heckj:
- update docs ordering to resolve swift-format lint concerns
- adding deprecation summary to redirect readers to correct API for deprecated versions
- updating .swift-format to preserve checking going forward

resolves #182

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -9,8 +9,7 @@ let package = Package(
         .library(name: "InMemoryTracing", targets: ["InMemoryTracing"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-service-context.git", from: "1.1.0"),
-        .package(url: "https://github.com/swiftlang/swift-docc-plugin", from: "1.0.0"),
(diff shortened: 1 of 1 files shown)


heckj:
The last commit in this changes a a number of access levels for properties and types in the Test module. Applying the documentation check applies broadly where-ever the swift-format lint command is triggered, and the current [GitHub-workflows script for soundness](https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/scripts/check-swift-format.sh) looks for _all_ .swift files in the repo, and doesn't distinguish between Tests and Sources.

I *think* that's a reasonable adjustment to support the existing moving parts, but happy to revisit and pull off that last commit if that's not desired.

Issue #189, opened 2025-10-04
NoOpTracer still sets up attributes when calling Span.updateAttributes


adam-fowler:
When you call `Span.updateAttributes(_:)` on a `NoOpSpan` the included closure shouldn't be called. This is the cause of unexpected allocations when the tracer has been set to NoOp.

Is adding the function to the protocol a breaking change if you provide a default implementation?

Not calling the closure could be considered a breaking change I suppose given people might do other stuff inside that closure.

Ideally SpanAttributes would be an associatedtype and the NoOpSpan would have its own version that didn't do anything, but I think that ship may have sailed.

kukushechkin:
What do you think of this design: add `attributeValue(_:)` as a new `Span` protocol requirement, and move `updateAttributes(_:)` from an extension to the protocol, keeping its current default implementation calling `update` unconditionally? This would let `NoOpSpan` override `updateAttributes(_:)` to skip the closure entirely, and let any other `Span` conformer control the attributes read/write if needed.

PR #203, opened 2025-12-18
Add updateAttributes as Span protocol requirement

adam-fowler:
Add `updateAttributes(_:)` as Span protocol requirement
The majority of Span implementations will use the default implementation
Adding the function to the protocol though allows us to override the function for the `NoOpSpan`

Fix for #189

### DIFF
--- a/Sources/Tracing/NoOpTracer.swift
+++ b/Sources/Tracing/NoOpTracer.swift
@@ -139,9 +139,8 @@ public struct NoOpTracer: LegacyTracer {
             }
         }
 
-
         /// Update the span attributes in a block instead of individually.
-        /// 
+        ///
         /// The NoOpSpan implementation does not call the closure as setting attributes
         /// on a NoOpSpan does nothing.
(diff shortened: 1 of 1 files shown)


ktoso (approved):
LGTM, this works fine because the default impl already exists; so this will be source compatible, which is the only guarantee we need here;

Therefore we can disregard the API break warning, it doesn't understand the default impl

Sources/Tracing/NoOpTracer.swift
```diff
@@ -139,6 +139,12 @@ public struct NoOpTracer: LegacyTracer {
             }
         }
 
+        /// Update the span attributes in a block instead of individually.
+        ///
+        /// The NoOpSpan implementation does not call the closure as setting attributes
+        /// on a NoOpSpan does nothing.
+        public func updateAttributes(_ update: (inout SpanAttributes) -> Void) {}
```
czechboy0: Hmm see my other comment, I still think we should call the closure, just with local empty span attributes that get discarded immediately, not saved on the instance.

PR #205, opened 2026-01-11
Bump Dependencies

bwdmr:
Bump on swift-service-context

Bump Docker files

Use Swift 6.0 throughout and bump dependencies on related packages.


kukushechkin:
I appreciate you taking the time to open this PR! However, we've just removed these files 
as they were significantly outdated. Since they're no longer in the codebase, this PR 
won't be needed. I'm closing it, but thanks for the heads up, sometimes this actually 
helps surface files that should be cleaned up.

PR #230, opened 2026-06-29
[SDT-0001]: task-local instrument implementation

### LINKED ISSUE #168: Task local tracer
Like https://github.com/apple/swift-log/issues/315, but for the tracer itself.

Would make testing easier.

kukushechkin:
Add `withTracer(_:_:)` for a task-local tracer scoping.

## Motivation:

`InstrumentationSystem.bootstrap(_:)` installs an instrument once per process, so every `withSpan`/`startSpan` and every read of `InstrumentationSystem.instrument` resolves through that single instrument. That fits one tracer for the whole application, but it leaves no way to bind a different instrument for a region of work — for example in tests, where a second `bootstrap` crashes and parallel tests cannot each install their own in-memory tracer.

## Modifications:

- Added `withTracer(_:_:)` and an internal `@TaskLocal _taskLocalInstrument`.
- A "Scoping a tracer using `withTracer`" guide.

## Result:

`swift-distributed-tracing` provides opt-in `withTracer(_:_:)` task-local tracer scoping. SDT-0001 proposal is implemented.

Closes #168


Sources/Instrumentation/WithInstrument.swift
```diff
@@ -0,0 +1,118 @@
...
+/// > ``InstrumentationSystem/instrument`` and `withSpan` / `startSpan`, which already observe whatever is
+/// > active.
+///
+/// - Parameters:
+///   - instrument: The instrument to make active for the duration of `operation`.
+///   - operation: The closure to run with `instrument` active.
+/// - Returns: The value returned by the closure.
+@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+@inlinable
+public func withInstrument<Result, Failure: Error>(
```
czechboy0: I wonder if it'd be nice to _also_ have `withTracer` variants, as most people, when they use SDT, instantiate and talk about "tracers", not "instruments".

If I think to how I'd discover this with autocompletion, unless intimately familiar with SDT, I'd struggle to realize that `withInstrument` is what I should use when I'm looking for `withTracer` (after I've already adopted `withLogger` and `withMetricsFactory`.

Sources/Tracing/Docs.docc/Proposals/SDT-0001-task-local-instrument.md
```diff
@@ -0,0 +1,170 @@
...
+public func withTracer<Result, Failure: Error>(
+    _ tracer: any Tracer,
+    isolation: isolated (any Actor)? = #isolation,
+    _ operation: () async throws(Failure) -> Result
+) async throws(Failure) -> Result
+#endif
+````
+
+The scope is backed by an `@TaskLocal` on ``InstrumentationSystem`` typed `(any Instrument)?`. There is no
+wrapper instrument, and nothing is installed on first use.
```
czechboy0: This is reasonable, have you also explored storing this in the ServiceContext?

It'd also be great to add some code examples of how people will bridge from manual passing to task local, and vice versa. I think that'll surface the repetition of having to manually propagate both the service context and the tracer, separately.

kukushechkin: > This is reasonable, have you also explored storing this in the ServiceContext?

Yeah. So this is a question of 2 task-local lookups vs 1 task-local lookup + ServiceContext `dict` lookup. The dictionary lookup is much slower, like 50% slower.

> It'd also be great to add some code examples of how people will bridge from manual passing to task local, and vice versa.

Do you think this should be right in the proposal, or, for example, in the instrumentation guide?

czechboy0: I think we should be forced to think about the impact on users, so I'd like to see it in the proposal.

Personally I think having to manually propagate two values instead of one is problematic enough to justify even slightly slower performance.

We need to consider the cumulative time all developers in the ecosystem will burn on debugging why their tracer got propagated and their service context didn't, or vice versa. We can make a decision here to completely design away that problem.

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -20,6 +20,10 @@ import ServiceContextModule
 ///
 /// Set up the instrumentation using ``bootstrap(_:)``, and access the globally available instrument using ``instrument``.
 /// If you need to use more that one cross-cutting tool you can do so by using ``MultiplexInstrument``.
+///
+/// To override the active instrument for a scope, such as a test or a subsystem, without touching the
+/// process-wide bootstrap, the Tracing module provides `withTracer(_:_:)`. It binds a task-local override
+/// that ``instrument`` and discovery check first, for the duration of a closure.
```
czechboy0: What does "discovery" mean here? Should it say "that ``instrument`` and `tracer` check first"?

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -53,21 +57,54 @@ public enum InstrumentationSystem {
...
+    /// ``instrument`` and ``_findInstrument(where:)`` check this first, for the duration of a scope, before
+    /// falling back to the bootstrapped instrument.
+    @TaskLocal
+    @usableFromInline
+    internal static var _taskLocalInstrument: (any Instrument)?
+
+    /// Runs `operation` with `instrument` bound to the task-local override. Backs `withTracer(_:_:)` in the
+    /// Tracing module. Marked `package`, not `internal`, so that module can call it.
+    @usableFromInline
+    package static func withTaskLocalInstrument<Result>(
```
czechboy0, line 74: Do we need these 3 methods? They're a very shallow abstraction. Could we just make `_taskLocalInstrument` visible at `package` level?

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -84,18 +121,47 @@ public enum InstrumentationSystem {
...
 
-    /// Returns the globally configured instrument.
+    /// The currently active ``Instrument``.
+    ///
+    /// This is the instrument bound by the innermost enclosing `withTracer(_:_:)` scope, if any,
+    /// otherwise the one set with ``bootstrap(_:)``. Returns a ``NoOpInstrument`` if neither was set.
     ///
-    /// Defaults to a no-op ``Instrument`` if ``bootstrap(_:)`` wasn't called before.
+    /// Checks the task-local first, and only reads the bootstrapped instrument if it is not set. A resolution
+    /// inside `withTracer(_:_:)` returns from the task-local directly and never touches the bootstrapped
```
czechboy0:
```suggestion
    /// inside `withTracer(_:_:)` returns from the task-local directly and never reads the bootstrapped
```

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -84,18 +121,47 @@ public enum InstrumentationSystem {
...
+    /// storage. An application that only calls ``bootstrap(_:)`` pays for that extra check on every
```
czechboy0: What should an adopter do with this last sentence? Is this a recommendation _not_ to use process bootstrap? Is it a performance FYI?

Sources/Instrumentation/InstrumentationSystem.swift
```diff
@@ -84,18 +121,47 @@ public enum InstrumentationSystem {
...
 }
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)  // for TaskLocal ServiceContext
 extension InstrumentationSystem {
     /// INTERNAL API: Do Not Use
+    ///
+    /// Finds the first instrument matching `predicate` in the currently active instrument. This is the
+    /// `withTracer(_:_:)` override if one is in scope, otherwise the bootstrapped instrument. If the
+    /// active instrument is a ``MultiplexInstrument``, its direct members are checked in order.
     public static func _findInstrument(where predicate: (Instrument) -> Bool) -> Instrument? {
```
czechboy0, line 132: Does this still need to be public? Who's using it outside this module? If only Tracing, could we make it `package`?

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -369,8 +369,42 @@ which is equivalent to surrounding the `withSpan` call with a binding of the con
 
 ### Global vs. "Stored" Tracers and Instruments
 
-Tracing works similarly to swift-log and swift-metrics, in the sense that there is a global "backend" configured at application start, by end-users (developers) of an application. And this is how using `InstrumentationSystem/tracer` gets the "right" tracer at runtime.
+Tracing works similarly to swift-log and swift-metrics, in the sense that there is a global "backend" configured at application start, by end-users (developers) of an application. And this is how using `Tracing/InstrumentationSystem/tracer` gets the "right" tracer at runtime.
```
czechboy0:
```suggestion
Tracing works similarly to Swift Log and Swift Metrics, in the sense that there is a global "backend" configured at application start, by end-users (developers) of an application. And this is how using `Tracing/InstrumentationSystem/tracer` gets the "right" tracer at runtime.
```

czechboy0 (commented):
Posting this partial review now, will continue review later.

czechboy0 (commented):
Posting this partial review now, will continue later.

Sources/Tracing/WithTracer.swift
```diff
@@ -0,0 +1,156 @@
...
+/// }
+/// ```
+///
+/// A `Tracer` is also an `Instrument`, so this replaces propagation too, not just span creation. To keep
+/// several tools active at once, install a `MultiplexInstrument` at `InstrumentationSystem/bootstrap(_:)`.
+///
+/// - Parameters:
+///   - tracer: The tracer to make active for the duration of `operation`.
+///   - operation: The closure to run with `tracer` active.
+/// - Returns: The value returned by the closure.@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
```
heckj:
```suggestion
/// - Returns: The value returned by the closure.
@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
```

`@available` needs to be on it's own line - I'm guessing an editing error somewhere in there ;-)

Sources/Tracing/WithTracer.swift
```diff
@@ -0,0 +1,156 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Distributed Tracing project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Makes `tracer` the active instrument for the current task and the child tasks it spawns, for the
+/// duration of `operation`.
+///
```
heckj:
```suggestion
/// Makes the tracer the active instrument for the current task and any child tasks it spawns.
///
/// The task-local tracer exists for the direction of `operation`.
```

^ This makes the abstract "plain" narrative English and drops the code voice (monospaced font), which doesn't reliably reflect in all the places that an abstract may be presented.

This is the exact same abstract that exists down at line 62/107 (the conditional compilation extension), and ideally there's a clear distinction between when someone should use *this* version and the others - I'm not clear enough on which is preferred and when, but if we can incorporate that (concisely) into the abstract itself, that's awesome - if not, making some note in the overview content beneath would be really helpful.

Sources/Instrumentation/InstrumentationSystem.swift
kukushechkin: No, `@TaskLocal` only propagates `@usableFromInline` [starting with 6.4](https://github.com/swiftlang/swift/issues/90094).

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -369,8 +369,42 @@ which is equivalent to surrounding the `withSpan` call with a binding of the con
...
 
-You may be tempted to allow users _configuring_ a tracer as part of your applications initialization. Generally we advice against that pattern, because it makes it confusing which library needs to be configured, how, and where -- and if libraries are composed, perhaps the setting is not available to the actual "end-user" anymore.
+#### Should my library accept a tracer parameter?
```
heckj: Rather than a rhetorical question, can we reframe this to be direct advice - "have your library accept a tracer parameter when..."

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -369,8 +369,42 @@ which is equivalent to surrounding the `withSpan` call with a binding of the con
...
+You may be tempted to allow users to configure a tracer as part of your applications initialization. Generally we advise against that pattern, because it makes it confusing which library needs to be configured, how, and where -- and if libraries are composed, perhaps the setting is not available to the actual "end-user" anymore.
+
+On the other hand, it may be valuable for testing scenarios to be able to set a tracer on a specific instance of your library. Therefore, if you really want to offer a configurable `Instrument` or `Tracer` then we suggest defaulting this setting to `nil`, and if it is `nil`, reaching to the global `InstrumentationSystem/instrument` or `Tracing/InstrumentationSystem/tracer` - this way it is possible to override a tracer for testing on a per-instance basis, but the default mode of operation that end-users expect from libraries remains working.
+
+#### Honoring a caller's scope
```
heckj: ideally the section title should be something directly actionable - maybe
```suggestion
#### Honor the caller's scope
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -369,8 +369,42 @@ which is equivalent to surrounding the `withSpan` call with a binding of the con
...
+
+A library observes a caller's ``withTracer(_:_:)-mixl`` scope only if it resolves the instrument *per call*,
+through the free-function `withSpan` / `startSpan` or `InstrumentationSystem/instrument` at the emission
+site. A library that captures a `Tracer` once at construction ignores any scope entered later, so resolve per
+call unless you deliberately want to pin one instrument to an instance.
+
+#### Testing your library's instrumentation
```
heckj:
```suggestion
#### Test your library's instrumentation
```

(making the heading direct and imperative)

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -438,6 +459,55 @@ Events usually show up in a in a trace view as points on the timeline (note that
 
 Events cannot be "failed" or "successful", that is a property of a ``Span``, and they do not have anything that would be equivalent to a log level. When a trace span is recorded and collected, so will all events related to it. In that sense, events are different from log statements, because one can easily change a logger to include the "debug level" log statements, but technically no such concept exists for events (although you could simulate it with attributes).
 
+### Scoping a tracer using withTracer
```
heckj:
```suggestion
### Scope a tracer using withTracer
```

(making the section heading more directly actionable voice)

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -438,6 +459,55 @@ Events usually show up in a in a trace view as points on the timeline (note that
...
+    }
+    #expect(tracer.finishedSpans.count == 1)
+}
+```
+
+Scoping the whole application is possible, but be aware that work that escapes the closure's task tree (a
+`Task.detached`, an `EventLoopFuture` callback, a long-lived background task) does not inherit the scope and
+falls back to the bootstrapped instrument, so a whole-application `withTracer` can silently misroute spans
+and propagated carrier data to the bootstrapped instrument at exactly those boundaries.
+
```
heckj:
```suggestion
It is possible to scope the whole application, but work that escapes the closure's task tree, such as a
`Task.detached`, an `EventLoopFuture` callback, or a long-lived background task, does not inherit the scope and
falls back to the bootstrapped instrument.
A whole-application `withTracer` can silently misroute spans
and propagate carrier data to the bootstrapped instrument at exactly those boundaries.
```

reworked to remove the parenthesis and break up the mega-sentence, and make the content more direct

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -369,8 +369,42 @@ which is equivalent to surrounding the `withSpan` call with a binding of the con
...
+#### Should my library accept a tracer parameter?
 
-On the other hand, it may be valuable for testing scenarios to be able to set a tracer on a specific instance of your library. Therefore, if you really want to offer a configurable `Instrument` or `Tracer` then we suggest defaulting this setting to `nil`, and if it is `nil`, reaching to the global `InstrumentationSystem/instrument` or `InstrumentationSystem/tracer` - this way it is possible to override a tracer for testing on a per-instance basis, but the default mode of operation that end-users expect from libraries remains working.
+You may be tempted to allow users to configure a tracer as part of your applications initialization. Generally we advise against that pattern, because it makes it confusing which library needs to be configured, how, and where -- and if libraries are composed, perhaps the setting is not available to the actual "end-user" anymore.
```
heckj:
```suggestion
You may be tempted to allow users to configure a tracer as part of your application's initialization. Generally we advise against that pattern, because it makes it confusing which library needs to be configured, how, and where --- and if libraries are composed, perhaps the setting is not available to the actual "end-user" anymore.
```

(switch to an em-dash in markdown, add missing apostrophe)

Sources/Instrumentation/InstrumentationSystem.swift
kukushechkin: It seems to be required by the `public static var tracer` and `legacyTracer` from `InstrumentationSystem`. This public symbol is not added in this PR.

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -369,8 +369,42 @@ which is equivalent to surrounding the `withSpan` call with a binding of the con
...
+
+On the other hand, it may be valuable for testing scenarios to be able to set a tracer on a specific instance of your library. Therefore, if you really want to offer a configurable `Instrument` or `Tracer` then we suggest defaulting this setting to `nil`, and if it is `nil`, reaching to the global `InstrumentationSystem/instrument` or `Tracing/InstrumentationSystem/tracer` - this way it is possible to override a tracer for testing on a per-instance basis, but the default mode of operation that end-users expect from libraries remains working.
+
+#### Honoring a caller's scope
+
+A library observes a caller's ``withTracer(_:_:)-mixl`` scope only if it resolves the instrument *per call*,
```
heckj:
```suggestion
A library observes a caller's ``withTracer(_:_:)-mixl`` scope only if it resolves the instrument _per call_,
```

(just making the markup consistent  - totally an OCD nit)

Sources/Instrumentation/InstrumentationSystem.swift
kukushechkin: "pays" == "pays for every task-local check", this is an explicit note on the internal resolution logic. Can be rephrased t be much more clear.

Sources/Instrumentation/InstrumentationSystem.swift
czechboy0: Can you elaborate? I'm not sure I follow.

Sources/Instrumentation/InstrumentationSystem.swift
czechboy0: Yeah I know, just since we're in this space I figured I'd ask the question. So does it still need to be public now?

kukushechkin:
@heckj @czechboy0 I've shortened most of the text, as @czechboy0  suggested and applied your suggestions to the edited version.

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -15,6 +15,14 @@ The first step to get metadata propagation and tracing working in your applicati
 A complete [list of swift-distributed-tracing implementations](http://github.com/apple/swift-distributed-tracing) 
 is available in this project's README. Select an implementation you'd like to use and follow its bootstrap steps.
 
+There are two ways to set up instrumentation:
+
+- `InstrumentationSystem/bootstrap(_:)`: process-wide, called once at startup. Simplest for a single tracer
+  used for the app's whole lifetime.
+- ``withTracer(_:_:)-mixl``: makes a tracer active for the current task while a closure runs, taking priority
```
czechboy0, line 20: I wonder if we should flip these, and recommend `withTracer` now. That'll reduce how often modern libraries fall back to the global one.

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -438,11 +454,46 @@ Events usually show up in a in a trace view as points on the timeline (note that
...
+### Scope a tracer using withTracer
+
+``withTracer(_:_:)-mixl`` makes a ``Tracer`` active for the current task while a closure runs, taking
+priority over the bootstrapped instrument. The binding propagates into child tasks, including an
+unstructured `Task { }`. It does not propagate into `Task.detached` or other work that escapes the closure's task
+tree, such as an `EventLoopFuture` callback or a long-lived background task. That work falls back to the
+bootstrapped instrument instead, so a whole-application `withTracer` call can silently misroute spans and
+carrier data at those boundaries.
+
+Use it for parallel-safe tests and per-subsystem overrides:
```
czechboy0: Maybe "Use it for all new code, and it's especially useful for parallel-safe tests and per-subsystem overrides"?

Sources/Tracing/Docs.docc/index.md
```diff
@@ -73,6 +73,6 @@ as the subsequent guides dive deeper into patterns and details of instrumenting
 - ``DefaultTracerClock``
 - ``NoOpTracer``
 
-### Contribute to the project
+### Proposal Process
```
czechboy0:
```suggestion
### Proposal process
```

README.md
```diff
@@ -53,4 +53,4 @@ If you know of any other library please send in a [pull request](https://github.
 
 ## Reference Documentation
 
-Please refer to the **[reference documentation](https://swiftpackageindex.com/apple/swift-distributed-tracing/documentation/tracing)** for detailed guides about adopting distributed tracing in your applications, libraries and frameworks.
+Please refer to the **[reference documentation](https://swiftpackageindex.com/apple/swift-distributed-tracing/documentation/tracing)** for detailed guides about adopting distributed tracing in your application's, libraries and frameworks.
```
czechboy0: I don't think this edit is right.

czechboy0 (changes requested):

FranzBusch (commented):
Overall LGTM except the comments from @czechboy0

README.md
kukushechkin: Oops

czechboy0 (approved):
Looks great, thank you @kukushechkin 🙏

heckj (approved):

PR #235, opened 2026-08-06
Fix MultiplexInstrument tracer fan-out

### LINKED ISSUE #232: `MultiplexInstrument` does not fan out to all tracers
`MultiplexInstrument` can hold multiple `Tracer`s, but only sends traces to one of them — is this expected?

The asymmetry is that `inject`/`extract` fan out to every member of the multiplex, but span creation doesn't: `InstrumentationSystem.tracer`/`legacyTracer` and the free-function `withSpan`/`startSpan` resolve to the first `Tracer` in the
multiplex and ignore the rest. So context is propagated through all members, but only one of them ever receives spans.

```
import Tracing
import InMemoryTracing

let tracerA = InMemoryTracer()
let tracerB = InMemoryTracer()
InstrumentationSystem.bootstrap(MultiplexInstrument([tracerA, tracerB]))

withSpan("request") { _ in }

print(tracerA.finishedSpans.count)  // 1
print(tracerB.finishedSpans.count)  // 0  ← tracerB never sees the span
```

So my questions:

Is "only the first `Tracer` receives spans" the intended contract? If so, it would be worth documenting on `MultiplexInstrument`, since "use multiple instruments" reads as though all of them are active. If not, should span creation fan out to every `Tracer` in the multiplex?

aryansk:
Fixes #232

MultiplexInstrument now forwards span creation and span lifecycle operations to every configured tracer, including context propagation, attributes, events, errors, links, status, flush, and end. This adds direct and global instrumentation coverage for multiple tracers.

Validation:
swift test
git diff --check

### DIFF
--- a/Sources/Tracing/InstrumentationSystem+Tracing.swift
+++ b/Sources/Tracing/InstrumentationSystem+Tracing.swift
@@ -16,13 +16,30 @@
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 extension InstrumentationSystem {
+    private static var multiplexTracer: MultiplexTracer? {
+        guard let multiplex = self.instrument as? MultiplexInstrument else { return nil }
+
+        let tracers =
+            multiplex
+            .matchingInstruments(where: { $0 is any LegacyTracer })
+            .compactMap { $0 as? any LegacyTracer }
+
+        guard tracers.count > 1 else { return nil }
+        return MultiplexTracer(tracers)
+    }
+
     /// Returns the ``Tracer`` bootstrapped as part of the `InstrumentationSystem`.
     ///
-    /// If the system was bootstrapped with a `MultiplexInstrument` this function attempts to locate the _first_
-    /// tracing instrument as passed to the multiplex instrument. If none is found, a ``NoOpTracer`` is returned.
+    /// If the system was bootstrapped with a `MultiplexInstrument` containing multiple tracing instruments, this
+    /// function returns a tracer that forwards span operations to each of them. If none is found, a ``NoOpTracer``
+    /// is returned.
     ///
     /// - Returns: A ``Tracer`` if the system was bootstrapped with one, and ``NoOpTracer`` otherwise.
     public static var tracer: any Tracer {
+        if let multiplexTracer = self.multiplexTracer {
+            return multiplexTracer
+        }
+
         let found: (any Tracer)? =
             (self._findInstrument(where: { $0 is (any Tracer) }) as? (any Tracer))
         return found ?? NoOpTracer()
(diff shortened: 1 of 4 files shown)


Sources/Tracing/InstrumentationSystem+Tracing.swift
```diff
@@ -16,26 +16,48 @@
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 extension InstrumentationSystem {
+    private static var multiplexTracer: MultiplexTracer? {
```
FranzBusch, line 19: Is this really necessary? A multiplex tracer or instrument should really just behave as any other tracer/instrument and should not need any special handling. It should be full on possible to implement the multiplex instrument/tracer outside of this module and the `instrument/tracer` properties behave correct.

aryansk: Thanks for raising this. The reason I put the forwarding type here is that `InstrumentationSystem.tracer` currently selects only the first tracer when bootstrapped with `MultiplexInstrument`; without an adapter at this boundary, global tracing calls do not fan out to the other configured tracers. I agree that a multiplex tracer could be implemented outside this module, so I don’t want to assume this is the preferred API boundary. Would you prefer keeping the global accessor behavior unchanged and moving fan-out to an external integration, or should this PR keep the global accessor behavior and add the smallest supported adapter? I’m happy to narrow the patch to the project’s preferred direction.

FranzBusch: Let me clarify. There can only be a single bootstrapped instrument or tracer. The fact that the current implementation checks if that bootstrapped instrument/tracer is a multiplex instrument is in my opinion wrong. Multiplex instruments/tracers are just as a regular instrument/tracer and they must not get any special treatment inside the core library here.

aryansk: Understood. I agree that the core tracing accessor should treat a bootstrapped multiplex value like any other tracer and should not inspect `MultiplexInstrument` specially. The current patch does not meet that boundary: it adds the special-case in `InstrumentationSystem+Tracing.swift`.

I will not push a speculative API rewrite. `MultiplexInstrument` lives in the Instrumentation target while `Tracer` and `Span` live in Tracing, so making the multiplex value a regular `Tracer` requires choosing the supported ownership/API boundary for that wrapper. Please confirm whether you want the multiplex tracer abstraction moved into the Instrumentation target or exposed as a separate tracing bootstrap value; once that boundary is confirmed, I can remove the special-case and update the tests accordingly.

Sources/Tracing/MultiplexTracer.swift
```diff
@@ -0,0 +1,182 @@
...
+package final class MultiplexSpan: Span, @unchecked Sendable {
+    private let spans: [any Span]
+
+    package init(_ spans: [any Span]) {
+        precondition(!spans.isEmpty, "A multiplex span must contain at least one span")
+        self.spans = spans
+    }
+
+    package var context: ServiceContext {
+        self.spans[0].context
```
kukushechkin, line 27: I think this needs to be discussed. Returning just the first context is at least confusing and needs documenting. But I would discuss what should be the expected context for a `MultiplexSpan`.

aryansk:
Agreed — returning the first context is an ambiguous contract, and `MultiplexSpan` needs an explicit decision before I change the implementation. I will hold this PR at the design boundary until maintainers decide whether the expected result is the first context, a documented merged/selected context, or another supported value; then I can update the implementation and regression tests to match.

Sources/Tracing/InstrumentationSystem+Tracing.swift
aryansk: Understood, thanks for clarifying. I agree the core accessor should treat a bootstrapped multiplex value like any other tracer and must not special-case `MultiplexInstrument`. I’ll leave this draft unchanged rather than move the abstraction across targets speculatively; once maintainers choose the supported ownership/API boundary, I can remove the special case and update the tests.

PR #239, opened 2026-09-09
[SDT-0002]: move ServiceContext from SSC to SDT proposal

kukushechkin:
Move `ServiceContext` and its supporting types out of the swift-service-context package into
a new `ContextStorage` target inside swift-distributed-tracing.

### Motivation

Most adopters reach `ServiceContextModule` only through `Tracing`, not as a direct dependency, so the
two-package split mostly adds complexity.

`ServiceContext` also reads as Swift Server specific. In practice it carries any ambient, propagated state,
not just service-to-service request state, and the name has stopped people from adopting
swift-distributed-tracing outside server applications.


FranzBusch (approved):
Let's run it!

PR #240, opened 2026-09-09
[SDT-0002]: move ServiceContext from SSC to SDT implementation

kukushechkin:


FranzBusch (approved):

czechboy0 (commented):
Did we consider putting the new types into one of the existing modules? Why do we want it in a separate one?

heckj:
@kukushechkin would you mind tacking on `ContextStorage` to the .spi.yaml as well, so that this module is included in SPI's documentation rendering.

```diff
diff --git a/.spi.yml b/.spi.yml
index 26f45b2..ea570cd 100644
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,4 +1,4 @@
 version: 1
 builder:
   configs:
-    - documentation_targets: [Tracing, Instrumentation, InMemoryTracing]
+    - documentation_targets: [Tracing, Instrumentation, InMemoryTracing, ContextStorage]
```

Sources/ContextStorage/TracingContext.swift
```diff
@@ -0,0 +1,296 @@
...
+///     } set {
+///       self[TestIDKey.self] = newValue
+///     }
+///   }
+/// }
+/// ```
+///
+/// For consistency, name key types with a `...Key` suffix (for example, `SomethingKey`) and the property
+/// used to access a value identifier by such key the prefix of the key (for example `something`).
+/// Please also observe the usual Swift naming conventions, such as prefering `ID` to `Id`, and so on.
```
heckj:
```suggestion
/// Please also observe the usual Swift naming conventions, such as preferring `ID` to `Id`, and so on.
```

Sources/ContextStorage/TracingContext.swift
```diff
@@ -0,0 +1,296 @@
...
+///
+/// ### Usage
+/// Using a context container is fairly straight forward, as it boils down to using the prepared computed properties:
```
heckj:
```suggestion
/// Using a context container is fairly straightforward, as it boils down to using the prepared computed properties:
```

Sources/ContextStorage/TracingContextKey.swift
```diff
@@ -0,0 +1,118 @@
...
+        self._nameOverride = keyType.nameOverride
+    }
+}
+
+extension AnyTracingContextKey: Hashable {
+    /// A Boolean value that indicates whether two service context keys are equivalent.
+    /// - Parameters:
+    ///   - lhs: The first service context key.
+    ///   - rhs: The second service context key.
+    /// - Returns: `True` if equivalent; otherwise `false`.
```
heckj:
```suggestion
    /// - Returns: `true` if equivalent; otherwise `false`.
```

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -272,7 +272,7 @@ func handler(request: HTTPRequest) async {
 }
 ```
 
-This is introducing multiple layers of nesting, and we have un-necessarily restored, picked-up, and restored the context again. In order to avoid this duplicate work, it is beneficial to use the ``withSpan(_:context:ofKind:at:function:file:line:_:)-8gw3v`` overload, which also accepts a `ServiceContext` as parameter, rather than picking it up from the task-local value:
+This is introducing multiple layers of nesting, and we have un-necessarily restored, picked-up, and restored the context again. In order to avoid this duplicate work, it is beneficial to use the ``withSpan(_:context:ofKind:at:function:file:line:_:)-8wgpb`` overload, which also accepts a `ServiceContext` as parameter, rather than picking it up from the task-local value:
```
heckj:
```suggestion
This is introducing multiple layers of nesting, and we have unnecessarily restored, picked-up, and restored the context again. In order to avoid this duplicate work, it is beneficial to use the ``withSpan(_:context:ofKind:at:function:file:line:_:)-8wgpb`` overload, which also accepts a `ServiceContext` as parameter, rather than picking it up from the task-local value:
```

Sources/Tracing/SpanProtocol.swift
```diff
@@ -12,16 +12,16 @@
...
 //===----------------------------------------------------------------------===//
 
-@_documentation(visibility: internal) @_exported import ServiceContextModule
+@_documentation(visibility: internal) @_exported import ContextStorage
 
 /// A span represents an interval from the start of an operation to its end, along with additional metadata included
 /// with it.
 ///
-/// A `Span` can be created from a `ServiceContext` or `LoggingContext` which MAY contain existing span identifiers,
+/// A `Span` can be created from a `TracingContext` or `LoggingContext` which MAY contain existing span identifiers,
```
heckj:
```suggestion
/// A `Span` can be created from a `TracingContext` which MAY contain existing span identifiers,
```

I didn't see a LoggingContext anywhere in this, so I think that was a copy/paste relic - double check me though.

Sources/ContextStorage/TracingContextKey.swift
```diff
@@ -0,0 +1,118 @@
...
+public protocol TracingContextKey: Sendable {
+    /// The type of value uniquely identified by this key.
+    associatedtype Value: Sendable
+
+    /// The human-readable name of this key.
+    ///
+    /// This name will be used instead of the type name when a value is printed.
+    ///
+    /// It MAY also be picked up by an instrument (from Swift Tracing) which serializes context items, such as used as
+    /// header name for carried metadata. Though generally speaking header names are NOT required to use the nameOverride,
```
heckj:
```suggestion
    /// It MAY also be picked up by an instrument (from Swift Tracing) which serializes context items, such as being used as the
    /// header name for carried metadata. Though generally speaking header names are NOT required to use the nameOverride,
```

mangled sentence - I *think* this was the intent?

PR #241, opened 2026-09-16
Docs update

heckj:
Docs cleanup
- removed the top level Docs/ directory since that's replicated into the sources now
- added some @_documentation() annotations to hide umbrella symbols added with `imported _export` from the documentation
- reset some of the abstracts to make overloads more clear - primarily withSpan, but also the protocol that offers different variants
- updated casing and cleaned up some redundancy in Tracing
- added InMemoryTesting as a module to host documentation through Swift Package Index


Sources/Instrumentation/Documentation.docc/Documentation.md
```diff
@@ -12,3 +12,7 @@ Base set of types which can be used to instrument libraries, excluding tracing s
 - ``Instrument``
 - ``Extractor``
 - ``Injector``
+
+### Supporting types
+
+- ``ReadWriteLock``
```
czechboy0: Wait is that type public?

slashmo: https://github.com/apple/swift-distributed-tracing/blob/main/Sources/Instrumentation/Locks.swift#L57 It is indeed. My best guess as to why is that we lacked `package`-level access back then, but even then we could have added an internal target to share it across `Instrumentation` and `Tracing`.

FranzBusch: Welp that was definitely a mistake :D

slashmo: Definitely. Hm, it used to be behind `@_spi(Locking)`.

czechboy0: I'd be for us moving it back under the SPI and asking for forgiveness if it breaks anyone.

Sources/Tracing/Docs.docc/Guides/InstrumentYourLibrary.md
```diff
@@ -1,5 +1,8 @@
 # Instrument Your Library or Framework
 
+Add distributed tracing support to the libraries and frameworks you maintain, so applications that
+depend on them get tracing for free.
```
kukushechkin: Not sure what is free in this context. I would try to highlight that application developers have more insights into what happens inside libs to better profile their applications.

```suggestion
depend on them can understand better what they do during profiling.
```

Sources/Tracing/Docs.docc/Guides/TraceYourApplication.md
```diff
@@ -1,5 +1,8 @@
 # Trace Your Application
 
+Instrument your server-side application with distributed tracing to gain visibility into its
```
kukushechkin: Should this be the point in time when we start to generalize the application scope of tracing APIs? :)

```suggestion
Instrument your application with distributed tracing to gain visibility into its
```

Sources/Tracing/Docs.docc/index.md
```diff
@@ -8,33 +8,34 @@ This is a collection of Swift libraries enabling the instrumentation of server s
...
-1. <doc:TraceYourApplication>, for **Application developers** who create server-side applications, 
-2. <doc:InstrumentYourLibrary>, for **Library/Framework developers** who provide building blocks to create these applications, 
-3. <doc:ImplementATracer>, for **Instrument developers** who provide tools to collect distributed metadata about your application.
+### Guides
 
-If unsure where to start, we recommend starting at the first guide and continue reading until satisfied, 
-as the subsequent guides dive deeper into patterns and details of instrumenting systems and building instruments yourself.
+We provide a number of guides aimed at getting started with tracing your systems, targeted at three
+different audiences: application developers, library and framework developers, and instrument
+implementers. See <doc:TraceYourApplication>, <doc:InstrumentYourLibrary>, and <doc:ImplementATracer>
```
kukushechkin: What do you think, should we reference them as developers as well?

```suggestion
different audiences: application developers, library and framework developers, and tracing instrument
developers. See <doc:TraceYourApplication>, <doc:InstrumentYourLibrary>, and <doc:ImplementATracer>
```

Sources/Tracing/TracerProtocol+Legacy.swift
```diff
@@ -131,7 +132,7 @@ extension LegacyTracer {
         )
     }
 
-    /// Start a new span returning an existential span reference.
+    /// Starts a new span using the default clock, returning an existential span.
```
kukushechkin: In the overload above the convenient default values are emphasized, should they be mentioned here as well?

```suggestion
    /// Starts a new span using the default clock and convenient default values for the
    /// remaining parameters, returning an existential span.
```

Sources/Instrumentation/Documentation.docc/Documentation.md
heckj: I'm totally cool with it - I suspect nobody is using it, I just curated it since it switched to public.

czechboy0: Can you slap that SPI on it, please, and remove from the doc page?

kukushechkin (approved):

